Issue #36, opened 2015-01-16
Spurios W5101 on abstract classes


mikebryant:
It doesn't make sense to require a `__unicode__` method on an abstract class.

carlio:
Fixed, or at least will be in the next release to PyPI.

Issue #37, opened 2015-01-24
Infinite loop calculating mro in astroid>=1.3.3


carlio:
As of astroid 1.3.3, class method resolution order is taken into consideration: https://bitbucket.org/logilab/astroid/commits/b1d8107715a4ceeeeb3f4a9065b67817e8cdc643

This caused an infinite loop when using pylint-django, as the model and form field transforms extended the base Django object then replaced it. This caused the MRO lookup to find a class as its own ancestor.

akshayjshah:
Any chance of cutting a new release with this fix?

carlio:
@akshayjshah I've been working on it but so far all my solutions have had the same problem. Getting closer though!

akshayjshah:
Gotcha - thanks!

carlio:
In the end I submitted a PR to astroid itself too: https://bitbucket.org/logilab/astroid/pull-request/67/mro-lookups-can-get-into-an-infinite-loop

However I don't know when that will be released so I've also patched pylint-django. Once I've verified it builds and the tests pass, and fixed a couple of small remaining bugs, I'll create a new PyPI release.

PR #38, opened 2015-01-24
No unicode warning on abstract models

carlio:
Addresses #36


carlio:
Oops, accidentally included fix for #37.

PR #45, opened 2015-04-04
added support for Django REST Framework

### LINKED ISSUE #44: old-style-class / Old-style class defined in Django REST Framework
I am using the ModelSerializer in Django REST Framework and pylint-django is marking the following code as old-style-class:

class MySerializer(serializers.ModelSerializer):
    class Meta:
        model = MyModel

lfrodrigues:
This adds support for DRF and fixes #44 
Had to update test suit a bit, not sure this was the best way but works.

### DIFF
--- a/pylint_django/augmentations/__init__.py
+++ b/pylint_django/augmentations/__init__.py
@@ -161,7 +161,8 @@ def is_model_meta_subclass(node):
 
     parents = ('django.db.models.base.Model',
                'django.forms.forms.Form',
-               'django.forms.models.ModelForm')
+               'django.forms.models.ModelForm',
+               'rest_framework.serializers.ModelSerializer')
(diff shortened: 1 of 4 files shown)


carlio:
Looks like coveralls wanted to be extra sure about this one :)

Thanks for the PR!

lfrodrigues:
Thanks for merging, when will you be making a release with this?

carlio:
I'll check this evening in case there are any small bugfixes I can include, then do the release tonight.

cancan101:
The exclude for `Meta` being an old school class should be on `BaseSerializer` no `ModelSerializer. BaseSerializers can have Meta too: https://github.com/tomchristie/django-rest-framework/blob/8c1365e315256f892dced348cf485bf966429596/rest_framework/serializers.py#L132

PR #47, opened 2015-04-20
Ensure pylint doesn't raise an error on 'raise SomeModel.DoesNotExist'

### LINKED ISSUE #46: Raising a DoesNotExist class gives E0702: Raising NoneType while only classes or instances are allowed (raising-bad-type)
If you have a model ```SomeModel``` and in your code you use ```raise SomeModel.DoesNotExist```, due to the transformation in ```pylint_django/transforms/transforms/django_db_models.py``` pylint raises E0702.

This code should raise no errors.

mikebryant:
Fixes #46

### DIFF
--- a/pylint_django/transforms/transforms/django_db_models.py
+++ b/pylint_django/transforms/transforms/django_db_models.py
@@ -1,3 +1,5 @@
+from django.core.exceptions import MultipleObjectsReturned, ObjectDoesNotExist
+
 class Model(object):
     _meta = None
     objects = None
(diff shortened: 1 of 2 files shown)


mikebryant:
False Positive? I didn't touch either of the files mentioned.

carlio:
Yeah don't worry about those. The first error has always been there but there were no tests for it until now, and the second looks like a bug in Landscape.

Thanks for the PR!

PR #48, opened 2015-05-25
Add support for more rest_framework classes, and django_filters

mbertolacci:
A previous PR added support for Django Rest Framework Meta classes, but only within serializers. Meta classes actually occur in many other places within code using DRF, so this PR expands this list to cover (hopefully) all of those.


carlio:
Sorry I completely forgot about this!

carlio:
This is included in `0.6.1`

PR #57, opened 2016-01-07
Fix travis builds

T045T:
The Travis build for Python2.6 was broken due to changes in astroid and deprecated options in pylint. I fixed both :)

PR #58, opened 2016-01-07
transforms: extend _locals instead of replacing it

T045T:
this way we don't clobber existing locals, something that the new test demonstrates.
previously, calls to save() and delete() on a model instance caused a pylint error.

The remaining travis build errors are handled in #57 - I'll update this one once that is merged :)


carlio:
Merged, after combining with #57 too, cheers!

PR #65, opened 2016-02-22
Add support of mongoengine module.

goldan:
Prevent 'no-member' errors for Document._meta, objects, id/pk, save and errors.

### DIFF
--- a/pylint_django/transforms/__init__.py
+++ b/pylint_django/transforms/__init__.py
@@ -39,3 +39,4 @@ _add_transform('django.db.models',
                'Model',
                'Manager')
 _add_transform('django.utils.translation', 'ugettext_lazy')
+_add_transform('mongoengine', 'Document')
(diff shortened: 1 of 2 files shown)


anentropic:
thanks for this, it's exactly what I needed!

mongoengine is not specific to Django though, this belongs in its own plugin rather than here IMHO

anentropic:
suggested change:  `objects = type`  (instead of `None`)

code like this is valid in mongoengine:
`uk_users = User.objects(country='uk')`

without the change above pylint complains that "objects is not callable"

nedbat:
Agreed that this good thing doesn't belong in a Django plugin.

goldan:
Agreed, though I don't feel like creating a new plugin for 15 lines of code... that's why I posted it here.

carlio:
Awesome stuff! Sorry for delay!

PR #67, opened 2016-03-07
Fix view.args and view.kwargs types

danni:
Args should be a tuple and kwargs a dict so that Pylint gets the types correct.


carlio:
Thanks :) sorry for delay merging, will look into a new release soon :)

Issue #72, opened 2016-06-03
ImportError: No module named pylint_django


erm0l0v:
Hi I got an error while try to run pylint with `pylint_django`

``` python
Traceback (most recent call last):
  File "/usr/local/bin/pylint", line 9, in <module>
    load_entry_point('pylint==1.4.1', 'console_scripts', 'pylint')()
  File "/Library/Python/2.7/site-packages/pylint/__init__.py", line 23, in run_pylint
    Run(sys.argv[1:])
  File "/Library/Python/2.7/site-packages/pylint/lint.py", line 1221, in __init__
    linter.load_plugin_modules(plugins)
  File "/Library/Python/2.7/site-packages/pylint/lint.py", line 450, in load_plugin_modules
    module = modutils.load_module_from_name(modname)
  File "/Library/Python/2.7/site-packages/astroid/modutils.py", line 137, in load_module_from_name
    return load_module_from_modpath(dotted_name.split('.'), path, use_sys)
  File "/Library/Python/2.7/site-packages/astroid/modutils.py", line 179, in load_module_from_modpath
    mp_file, mp_filename, mp_desc = imp.find_module(part, path)
ImportError: No module named pylint_django
```

`python -c "import pylint_django"` - works fine

-----

Versions:

OS: OS X 10.11.4
Python: 2.7.10
pylint: 1.5.5
pylint-django: 0.7.1

lezorich:
I have the same error!

carlio:
https://github.com/landscapeio/pylint-django/pull/73 will hopefully help with this, will close for now and make a release soon.

PR #73, opened 2016-06-03
#72 Add zip_safe

erm0l0v:
According to this issue https://github.com/landscapeio/prospector/issues/18 this change should help


carlio:
Yep that should hopefully fix a few strange path errors reported, no idea why but no harm in trying :)

PR #74, opened 2016-06-06
Supress no-member errors for inherited attributes

OscaTutenchamon:
Weather this is a problem of pylint not understanding inheritance or Django metaclasses magic I've found that pylint still reports many `no-member` errors when accessing attributes of custom classes inheriting from Django bases (Manager / ModelAdmin / etc.):

```python
from django.contrib import admin


class FooAdmin(admin.ModelAdmin):

    # ...
    readonly_fields = ('_readonly_foo',)

    def _readonly_foo(self, obj):
        if not obj.pk:
            return self.empty_value_display
        # ...

```
gives
```
Instance of 'FooAdmin' has no 'empty_value_display' member (no-member)
```
The same goes for some other classes.

This pull request supresses `no-member` error for attribute `X.attr` if `X` is  subclass of `Y` and `Y` defines attribute `attr`.

### DIFF
--- a/pylint_django/augmentations/__init__.py
+++ b/pylint_django/augmentations/__init__.py
@@ -6,6 +6,7 @@ from pylint.checkers.classes import ClassChecker
 from pylint.checkers.newstyle import NewStyleConflictChecker
 from pylint.checkers.variables import VariablesChecker
 from astroid import InferenceError
+from astroid.objects import Super
 from pylint_django.compat import ClassDef, ImportFrom, Attribute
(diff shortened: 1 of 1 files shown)


carlio:
Great stuff, thank you! One day I'll figure out a way to make it so that hard-coding so many attributes is not necessary. Will make a new release soon once I clear out the remaining issues.

OscaTutenchamon:
Hey, is new release coming any time soon? It's kinda ugly to have github link in requirements ;)
Also could you add me to contributors? I forget to do it and I think it would be awesome.

PR #93, opened 2017-04-21
Only return ClassDefs when inserting type shims into fields

cpennington:
I hit this bug when using freezegun with pylint-django (but am unable to really produce a reasonable reproduction case).


atodorov:
Hi @cpennington,
it's been a while but we're back on track with this project. 

Can you rebase & push again so that Travis CI builds get restarted ? 

Are you still able to reproduce this problem or at least let me know how to reproduce locally ?

cpennington:
Honestly, I haven't looked at it in a long time. I'm not sure if it still repros on my environment.

atodorov:
Closing since there doesn't seem to be a reproducer and this has probably been resolved meanwhile. We can always reopen if that starts to reproduce again.

PR #98, opened 2017-08-08
Using local copy of pylint.testutils if import fails

### LINKED ISSUE #97: Breaks under PyLint 1.7.x
Pylint 1.7 removes the `make_tests` function, causing tests to fail.

jamesmallen:
Fixes #97 

PyLint removed/made private many of the methods that were being used to generate tests. This PR addresses that by including a local copy of pylint's testutils.py and falling back to that.

See discussion here: https://github.com/PyCQA/pylint/issues/1438

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -1,2 +1,2 @@
 #!/bin/bash
-python test/test_func.py
+python -m test.test_func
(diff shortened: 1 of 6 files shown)


atodorov:

PR #99, opened 2017-08-08
Adding tests and transforms for DurationField

### LINKED ISSUE #95: Instance of 'DurationField' has no 'total_seconds' member
The `models.DurationField` (which behaves like a `datetime.timedelta`) added in Django 1.8 does not seem to be supported.

jamesmallen:
Fixes #95 

This PR adds support for [DurationField](https://docs.djangoproject.com/en/1.8/ref/models/fields/#durationfield), added in Django 1.8. DurationField acts like a `datetime.timedelta` object.

Minimal tests are included. Note that due to updates in PyLint, all tests fail, although tests pass after merging #98 into my local copy.


atodorov:

PR #111, opened 2018-01-08
Fixing NameChecker's const regex expression could be NoneType bug

leven-cn:
This bug appears in both Python 3.5 and 3.6


atodorov:

PR #113, opened 2018-01-18
Merge updated fork

atodorov:
Changes:

* make tests running again
* remove testing for Python 2 and older versions of Django (testing only with 1.11 and 2.0+ on Py3). No modifications to the code though, should still work on older versions.
* merged all other pending PRs, keeping authorship credentials so we can merge directly instead of going over each PR individually.

TODO: will remove custom naming and versioning coming from the fork and let you know when this is ready to merge and review.


carlio:
This all looks good to me!

Two comments - first, makes sense to wait for @colinhowe to have a look and also see what his changes are, may as well merge at the same time.

Second - it'd probably be good to have some kind of deprecation or warning maybe in `setup.py` that Py2 is not supported, as lots of people don't install a specific version so will get the latest and suddenly builds break (trust me on that one!). Does that make sense? Maybe just a sysexit with a printed line saying "if you want to use py2 with pylint-django please install a version <0.8". With better wording.

atodorov:
@carlio I can add the deprecation warning in a separate commit before we release 0.8. I have only removed testing for Python 2, mostly because I didn't have the time to deal with fixing the Python 2 tests. I have not modified the code to remove anything related to Py2 so everything will still work.

And we can make version 0.8.0 the last that will support Py2, then declare this project Py3 only and move on.

carlio:
> because I didn't have the time to deal with fixing the Python 2 
Oh I understand that completely!

Sounds all good to me.

colinhowe:
I'm happy with all of this. My changes have made it in already via the work that @atodorov has done

carlio:
Let's do it! I'll let you push the button to make sure permissions are set up right, let me know if there are any problems

PR #116, opened 2018-01-18
Don't transform DoesNotExist. Fixes #81

### LINKED ISSUE #81: duplicate-except throws red-herring on DoesNotExist
Hi guys, I have the following issue.
I raised it originally to the pylint team who thinks this is actually due to some processing you are doing with the DoesNotExist (I guess for other code checks)!

https://github.com/PyCQA/pylint/issues/1050
### Steps to reproduce
1. in a django project with 2 models
2. create a piece of code like this

```
    try:
        print("foo")
    except Model1.DoesNotExist:
        print("bar")
    except Model2.DoesNotExist:
        print("baz")
```
### Current behavior

This raises a W0705 duplicate-except saying it's catching django.core.exceptions.ObjectDoesNotExist. both those exception (as far as I understand it) are indeed inheriting from ObjectDoesNotExist (common ancestor), but they are not the same class
### Expected behavior

Do not raise this issue on common ancestor
### pylint --version output

pylint 1.5.6,
astroid 1.4.6
Python 2.7.12 (default, Jun 29 2016, 14:05:02)
[GCC 4.2.1 Compatible Apple LLVM 7.3.0 (clang-703.0.31)]

atodorov:
@gbataille, @sergafts can you try this patch and see if it works for you? 

@carlio from what I understand the transforms code adds a DoesNotExist attribute to the model class in such a way that it is the same type (ObjectDoesNotExist) which then triggers the original pylint warning reported in #81. I just have no idea if the fix won't break something else.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@
 * Update tests and require latest version of pylint (>=1.8), fixes
   [#53](https://github.com/landscapeio/pylint-django/issues/53),
   [#97](https://github.com/landscapeio/pylint-django/issues/97)
+* [#81](https://github.com/landscapeio/pylint-django/issues/81) Fix 'duplicate-except' false negative
+  for except blocks which catch the `DoesNotExist` exception.
 
 ## Version 0.7.4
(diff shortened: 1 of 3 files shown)


colinhowe:
This looks OK to me. If it breaks something then we need another test for whatever that is :)

PR #126, opened 2018-01-27
Fix flake8 complaints

bittner:
This PR addresses issue #102.

Issue #102, opened 2017-10-07
pylint_django does not pass flake8


bittner:
I know, this may sound a bit shady as a comment, but the project does not pass `flake8`. Not even when you ignore line length generously:

```bash
$ flake8 --max-line-length=120
./compat.py:5:5: F401 'astroid.nodes.Class as ClassDef' imported but unused
./compat.py:5:5: F401 'astroid.nodes.Function as FunctionDef' imported but unused
./compat.py:5:5: F401 'astroid.nodes.From as ImportFrom' imported but unused
./compat.py:5:5: F401 'astroid.nodes.AssName as AssignName' imported but unused
./compat.py:5:5: F401 'astroid.nodes.Getattr as Attribute' imported but unused
./compat.py:17:9: F401 'astroid.util.Uninferable' imported but unused
./plugin.py:10:1: F401 'pylint_django.transforms' imported but unused
./transforms/transforms/django_utils_translation.py:1:1: E731 do not assign a lambda expression, use a def
./transforms/transforms/mongoengine.py:4:1: E302 expected 2 blank lines, found 1
./transforms/transforms/mongoengine.py:14:5: E731 do not assign a lambda expression, use a def
./transforms/transforms/django_db_models.py:3:1: E302 expected 2 blank lines, found 1
./transforms/transforms/django_forms.py:14:5: E731 do not assign a lambda expression, use a def
./transforms/transforms/django_forms_fields.py:104:5: E303 too many blank lines (2)
./transforms/transforms/django_db_models_fields_files.py:12:1: W391 blank line at end of file
./augmentations/__init__.py:390:5: E265 block comment should start with '# '
./augmentations/__init__.py:391:5: E265 block comment should start with '# '
./checkers/__init__.py:9:60: W292 no newline at end of file
```

This may not be a big deal, but it's also something that's easily fixed.

Not sure though if you want "another tool" in this repository.

bittner:
Congratulations on the move to PyCQA!

Now it's time to clean up the code base.

atodorov:
@bittner care to contribute a pull request for the flake8 stuff ?

bittner:
I do care.

bittner:
Just a quick summary.

Results of PR #126 and #127:

- We now have made [most parts of the code base](https://travis-ci.org/PyCQA/pylint-django/jobs/338127371#L473) follow PEP8 and other rules enforced by flake8
[...]

PR #127, opened 2018-01-31
Add Tox configuration, use Tox on Travis-CI

bittner:
This PR adds a Tox configuration to the project to help keeping code quality stable.

The Travis setup is adapted accordingly: `tox` is used to run all linting and tests also on Travis CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,19 +1,33 @@
-sudo: false
 language: python
 python:
-  - "3.5"
-  - "3.6"
+  - 2.7
+  - 3.4
+  - 3.5
+  - 3.6
 env:
-  - DJANGO="1.11,<2.0"
-  - DJANGO="2.0,<3.0"
+  - DJANGO=1.11
+  - DJANGO=2.0
+matrix:
+  exclude:
+    # Python/Django combinations that aren't officially supported
+    - { python: 2.7, env: DJANGO=2.0 }
+  include:
+    - { python: 3.6, env: TOXENV=flake8 }
+    - { python: 3.6, env: TOXENV=pylint }
+    - { python: 3.6, env: TOXENV=readme }
+  allow_failures:
+    - env: TOXENV=flake8
+    - env: TOXENV=pylint
+    - python: 2.7
 
 install:
-  - pip install coverage coveralls djangorestframework psycopg2 django-tables2
-  - pip install pylint-plugin-utils "Django>=$DJANGO"
+  - pip install tox-travis
 script:
-  - PYTHONPATH=. coverage run test/test_func.py
+  - tox
+
 after_success:
-  coveralls
+  - pip install coveralls
+  - coveralls
 notifications:
   email:
     on_failure: change
--- a/tox.ini
+++ b/tox.ini
@@ -1,0 +1,47 @@
+# This configuration file is for Tox. https://tox.readthedocs.io/
+# It houses configuration sections for common Python tools, too.
+
+[tox]
+envlist =
+    flake8
+    pylint
+    readme
+    py{27,34,35,36}-django111
+    py{34,35,36}-django20
+
+[testenv]
+commands =
+    flake8: flake8
+    pylint: pylint --rcfile=tox.ini pylint_django setup
+    readme: python setup.py check --restructuredtext --strict
+    py{27,34,35,36}-django{111,20}: coverage run test/test_func.py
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -1,19 +1,43 @@
...
+  - 3.4
+  - 3.5
+  - 3.6
 env:
-  - DJANGO="1.11,<2.0"
-  - DJANGO="2.0,<3.0"
+  - DJANGO=1.7
+  - DJANGO=1.8
+  - DJANGO=1.9
+  - DJANGO=1.10
```
atodorov: Django versions before 1.11 are end-of-life so I suggest we don't test with them.

.travis.yml
```diff
@@ -1,19 +1,43 @@
...
+    # Python/Django combinations that aren't officially supported
+    - { python: 3.5, env: DJANGO=1.7 }
+    - { python: 3.6, env: DJANGO=1.7 }
+    - { python: 3.6, env: DJANGO=1.8 }
+    - { python: 3.3, env: DJANGO=1.9 }
+    - { python: 3.6, env: DJANGO=1.9 }
+    - { python: 3.3, env: DJANGO=1.10 }
+    - { python: 3.6, env: DJANGO=1.10 }
+    - { python: 3.3, env: DJANGO=1.11 }
+    - { python: 2.7, env: DJANGO=2.0 }
```
atodorov: If we don't test with older Django versions then you can "include" python 2.7 and Django 1.11 instead.

bittner:
Good hint, thanks! It should be okay now. Anything else?

tox.ini
```diff
@@ -0,0 +1,47 @@
...
+    pylint: Django>=2.0,<2.1
+    readme: readme_renderer
+    django111: Django>=1.11,<2.0
+    django20: Django>=2.0,<2.1
+    py{27,34,35,36}: coverage
+    py{27,34,35,36}: djangorestframework
+    py{27,34,35,36}: psycopg2
+    py{27,34,35,36}: django-tables2
+    py{27,34,35,36}: pylint-plugin-utils
+    py{27,34,35,36}: pytest
```
atodorov: Is it possible to add all of these into a requirements.txt (for devel) file and specify that in tox.ini ?

bittner: I don't think so. These are conditional dependencies. I think it's possible to use `-rrequirement.txt`, but we would then probably have to create a single requirements file for each condition.

atodorov:
LGTM right now, maybe squash the two commits together. I have no preference whether or not we place dependencies in requirements.txt now or in a separate PR (it at all possible). 

However I see failing tests. Python 2 + Django 1.11 is failing and you have to fix these before we can merge.

flake8 is also failing, maybe needs a rebase so it can pick up the related changes you made earlier. 
pylint is also failing, not sure about that. Maybe you should tell Travis this is allowed to fail if it is non trivial to fix. I', not comfortable merging a PR that will cause all subsequent builds to fail, even if it is just because of some pylint warnings.

bittner:
The point of this PR, apart from making checks run automatically, is to demonstrate that checks and tests are failing. This slips unnoticed when they are not executed automatically (as in PR #126).

I think I mentioned in PR #126 that I refrained from fixing some of flake8's complaints, because the result could be unsure. Someone with more knowledge of the inner workings of this software should continue with the adaptions - until all checks and tests are green.

tox.ini
atodorov: As it is now all of these dependencies are installed unconditionally in .travis.yml except accounting for the Django version. I'm fine to continue to do so because these dependencies are only used during testing.

atodorov:
> Someone with more knowledge of the inner workings of this software should continue with the adaptions - until all checks and tests are green.

I see your point. In this case let's instruct Travis to accept failures on the flake8 and pylint test jobs and I will try to fix some of them but can't tell if it will be this or next week.

bittner:
Can you squash the 3 commits here on GitHub (it's an option in the Merge settings), or should I do this manually?

atodorov:
> Can you squash the 3 commits here on GitHub (it's an option in the Merge settings), or should I do this manually?

better do it manually so I don't have to go creative with the commit log

bittner:
All done. I hope it's fine for you now.

bittner:
When this PR is merged we still need to keep issue #102 open.

bittner:
Thanks for merging!

Issue #96, opened 2017-07-10
make django a requirement, or somehow let users know that it has to be installed


jo-tham:
Hi,

thanks for an awesome plugin; it was a relief to find when I introduced pylint to an existing django project and was swarmed with errors from django `class Meta` and so forth.

My team had an issue with getting the plugin to work in some instances.

Specifically; using a virtualenv/global installation of pylint and plylint django which did not include django. It may seems silly, but I think it's a valid usage pattern.

The django typical lint errors kept being reported, and we were pretty confused as to why. Of course, django had to be installed on path.

I didn't see anywhere in the readme that django had to be installed. It's a minor nuisance to solve, but it'd be nice to help the next person by either:

* letting them know in the readme
* raising an error when pylint_django is loaded if django is not installed
* adding django to requirements in `setup.py`

I'm in favour of the last option. Glad to create a PR for the option that the maintainer(s) prefer, if any.

Thanks

carlio:
@jo-tham There is already a checker to do that here: https://github.com/landscapeio/pylint-django/blob/master/pylint_django/checkers/django_installed.py

So if you did not get the warning then that is probably broken! There have been many version of pylint since it was originally written so perhaps the API changed somewhere. I'll look into it :-)

atodorov:
With the latest version I get a ModuleNotFoundError: No module named 'django' exception if Django is not found in the path. I'm not sure the existing checker will work b/c Django is imported in other modules before the checker is registered. 

I am in favor of closing this issue and removing the django_installed.py checker. @carlio @colinhowe what do you think ?

carlio:
The checker was at least originally loaded before other things - it was there to avoid needing to specify Django as a requirement to avoid polluting peoples' environments as who knows what version or how they installed it. That checker is there really just as a hint for when things go wrong. It can go away for sure, it's not really important except to help quickly diagnose why pylint-django isn't working, it's not a code quality issue.

@atodorov  It'd be nice to keep that warning (since some CI environments might not have a full dependency tree) but the checker itself can go.

carlio:
Also https://github.com/PyCQA/pylint-django/blob/master/pylint_django/compat.py#L23 would need to be changed a bit!

atodorov:
@carlio I  can always try `import django` and exit gracefully before anything else and we'll not have to update the compat layer. Will give it a try later.

carlio:
[...]

PR #131, opened 2018-03-04
Adjust where tests are located

atodorov:
@moubctez,  @aerostitch,
I think this PR takes care of the test directory not being directly installed under `site-packages/` while still carrying all test related files in the tarball to accommodate Debian packaging. 

Let me know if you see anything I'm missing here. I can release 0.9.2 without further if the proposed changes work for you.


aerostitch:
@atodorov looks great. The only thing that is missing is the modification of `scripts/test.sh` I would say.

carlio:
👍  from me too

atodorov:
@aerostitch fixed `scripts/test.sh`, good catch. 

FYI I will do a quick fix for #96 as well and probably release the new version tomorrow.

PR #132, opened 2018-03-05
List Django as a dependency. Fix #96

### LINKED ISSUE #96: make django a requirement, or somehow let users know that it has to be installed

atodorov:
also add a big warning statement about possible side effects and to deter people from reporting issues about Django version mismatches. Updates README formatting a bit as well.

@jakirkham this should fix #96.

@carlio I've read your reasoning about not listing this as a dependency and the side effects it may have. Let's see if the warning text I've added has any effect. If we start seeing consistent reports about Django version mismatches I'm all for dropping the Django dependency from `setup.py` and adjusting the warning so that folks are aware they have to install Django explicitly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,19 +10,33 @@ pylint-django
 
 `pylint-django` is a [Pylint](http://pylint.org) plugin for improving code analysis for when analysing code using Django. It is also used by the [Prospector](https://github.com/landscapeio/prospector) tool.
 
-## Usage
+# Installation
 
-#### Pylint
+```
+pip install pylint-django
+```
 
-Ensure `pylint-django` is installed and on your path (`pip install pylint-django`), and then run pylint:
+**WARNING:** `pylint-django` requires `Django` to be installed. Our `setup.py` file doesn't
+specify which particular version of Django is going to be installed because we have no idea
+what version is used inside your project. The latest version of Django will be installed if
+it has not been installed beforehand! DO NOT report issues about mismatching Django versions
+if that happens. Instead get your testing environment sorted out and make sure that you have
+the appropriate version of Django installed!
+
+# Usage
+
+## Pylint
+
+Ensure `pylint-django` is installed and on your path and then execute:
 
 ```
 pylint --load-plugins pylint_django [..other options..]
 ```
 
-#### Prospector
+## Prospector
 
-If you have `prospector` installed, then `pylint-django` will already be installed as a dependency, and will be activated automatically if Django is detected.
+If you have `prospector` installed, then `pylint-django` will already be installed as a dependency,
+and will be activated automatically if Django is detected.
 
 ```
 prospector [..other options..]
@@ -42,14 +56,14 @@ Please feel free to add your name to the `CONTRIBUTORS.md` file if you want to b
 credited when pull requests get merged. You can also add to the `CHANGELOG.md` file
 if you wish, although I'll also do that when merging if not.
 
-## Tests
+# Tests
 
 The structure of the test package follows that from pylint itself.
 
 It is fairly simple: create a module starting with `func_` followed by
 a test name, and insert into it some code. The tests will run pylint
- against these modules. If the idea is that no messages now occur, then
- that is fine, just check to see if it works by running `scripts/test.sh`.
+against these modules. If the idea is that no messages now occur, then
+that is fine, just check to see if it works by running `scripts/test.sh`.
 
 Ideally, add some pylint error suppression messages to the file to prevent
 spurious warnings, since these are all tiny little modules not designed to
--- a/setup.py
+++ b/setup.py
@@ -17,6 +17,7 @@ setup(
     install_requires=[
         'pylint-plugin-utils>=0.2.1',
         'pylint>=1.8.2',
+        'Django',
     ],
     license='GPLv2',
     classifiers=[


JensTimmerman:
I don't really like this change.
I use prospector for a bunch of python projects that are not at all django related.
Since I use prospector pylint-django get's installed automatically, this was not an issue, until now, because pylint-django now tries to install Django as well. (version 2.x, which doesn't even work on python 2.7)
 So all my tests now try to install django, and fail.

Not sure if pylint-django is to blame, or prospector, but it seems at least the pylint-django developers are awere that pylint-django is auto installed with prospector...

README.md
```diff
@@ -10,19 +10,33 @@ pylint-django
...
 ```
 pylint --load-plugins pylint_django [..other options..]
 ```
 
-#### Prospector
+## Prospector
 
-If you have `prospector` installed, then `pylint-django` will already be installed as a dependency, and will be activated automatically if Django is detected.
+If you have `prospector` installed, then `pylint-django` will already be installed as a dependency,
+and will be activated automatically if Django is detected.
```
JensTimmerman: won't Django always be detected? Since you're adding it as a dependency here?

jakirkham:
This sounds like a poor separation of concerns, @JensTimmerman. Really prospector should soften the pylint-django requirement if it is not always needed. Have you brought this up with the prospector devs?

JensTimmerman:
@jakirkham see https://github.com/landscapeio/prospector/issues/245

carlio:
It's not a poor separation of concerns until recently - the point of prospector was to be a super easy "run one command" static analysis tool so it made sense to include pylint-django because automatically tuning checks for whatever framework you were using was the initial focus.

I realise now that the propagation of requirements was another reason I didn't want to include django as a requirement of pyling-django. By removing pylint-django as a default install for prospector, it undermines the whole point which was "install, point it at code, get results with as little effort as possible".

I still maintain that not including Django as a dependency of pylint-django is a good idea, people who don't set up their test environment properly need to deal with that and including as many warnings and clues as possible in pylint-django is the best idea - hence the DjangoNotInstalledChecker. Make it easy to diagnose but don't try to second guess every environment it will be installed into because that's too many permutations.

carlio:
(Also I should point out that the "prospector devs" is me and just me, and until recently that was true of pylint-django too, so... I am the main one to blame for faults in both 🤕 )

atodorov:
@carlio as a pylint-django developer and user I'm not really interested in prospector and I'm more in favor with @jakirkham that prospector needs to relax the pylint-django dependency. 

OTOH I recognize there's a valid use-case for prospector to pull in pylint-django automatically but not pull Django because, well you should already have Django. 

IMO both the fact that prospector has a dependency on pylint-django and the fact that I and maybe others were not aware of this lead to the conflict. @JensTimmerman brings up a valid use-case too, I don't argue about it. 

Can we all agree to revert the Django dependency from pylint-django and improve README where the fat warning will say that Django must be installed manually ? I can also add a GitHub issues template to warn users not to report bugs against that.

atodorov:
How about an optional dependency on Django? 

`pip install pylint-django[with-Django]`

plus the above suggestion for README ?

jakirkham:
Sorry the argument doesn't seem very persuasive to me.

One cannot even import `pylint-django` without `django` being installed. So it doesn't make sense to have `pylint-django` without this being a hard requirement. One could probably investigate softening the `django` requirement to the point that `pylint-django` installs and is importable without it (perhaps with all or nearly all functionality disabled). Not sure this is worth the effort, but will happily let `pylint-django` developers make that decision.

My recommendation would be that `prospector` soften the `pylint-django` dependency, but add it as extra requirement. That way one could do something like `pip install prospector[all]` and get `pylint-django` and any other cool, optional dependencies along with it. However if users want to install `prospector` without `pylint-django` that would be an option as well. Would think this added flexibility would make `prospector` a more appealing tool to users not less, but could be wrong about that.

atodorov:
I think there's a bug in the existing code base and the django_installed checker does not work. IMO this is the source of @jakirkham's arguments and probably some confusion too.

Here's the output on a brand new virtualenv without Django (pylint-django 0.9.1):
```
$ pylint --load-plugins=pylint_django ~/Kiwi/tcms
  File "/home/senko/.virtualenvs/testme/bin/pylint", line 11, in <module>
    sys.exit(run_pylint())
  File "/home/senko/.virtualenvs/testme/lib/python3.6/site-packages/pylint/__init__.py", line 16, in run_pylint
    Run(sys.argv[1:])
  File "/home/senko/.virtualenvs/testme/lib/python3.6/site-packages/pylint/lint.py", line 1268, in __init__
    linter.load_plugin_modules(self._plugins)
  File "/home/senko/.virtualenvs/testme/lib/python3.6/site-packages/pylint/lint.py", line 494, in load_plugin_modules
    module = modutils.load_module_from_name(modname)
  File "/home/senko/.virtualenvs/testme/lib/python3.6/site-packages/astroid/modutils.py", line 190, in load_module_from_name
    return load_module_from_modpath(dotted_name.split('.'), path, use_sys)
  File "/home/senko/.virtualenvs/testme/lib/python3.6/site-packages/astroid/modutils.py", line 233, in load_module_from_modpath
    module = imp.load_module(curname, mp_file, mp_filename, mp_desc)
  File "/home/senko/.virtualenvs/testme/lib64/python3.6/imp.py", line 245, in load_module
    return load_package(name, filename)
  File "/home/senko/.virtualenvs/testme/lib64/python3.6/imp.py", line 217, in load_package
    return _load(spec)
  File "<frozen importlib._bootstrap>", line 684, in _load
  File "<frozen importlib._bootstrap>", line 665, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 678, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/home/senko/.virtualenvs/testme/lib/python3.6/site-packages/pylint_django/__init__.py", line 7, in <module>
    from pylint_django import plugin
  File "/home/senko/.virtualenvs/testme/lib/python3.6/site-packages/pylint_django/plugin.py", line 5, in <module>
    from pylint_django.augmentations import apply_augmentations
  File "/home/senko/.virtualenvs/testme/lib/python3.6/site-packages/pylint_django/augmentations/__init__.py", line 19, in <module>
    from django.views.generic.base import View, RedirectView, ContextMixin
ModuleNotFoundError: No module named 'django'
```

PR #137, opened 2018-03-09
Multiple packaging fixes

atodorov:
There are 3 major changes here:

1) Revert the dependency on Django instroduced in #132 and convert it to optional dependency
2) Fix the DjangoInstalled checker to actually work
3) Add tests for all of the above plus a few more tests to build and sanity check the resulting packages.

@carlio , @JensTimmerman, @jakirkham please comment on this.


JensTimmerman:
looks good to me 👍

PR #143, opened 2018-03-23
Suppress not-an-iterable for model_utils.managers. Fixes #117

### LINKED ISSUE #117: Incorrectly identify queryset as not iterable when model manager is a subclass
Since upgrading to version 0.8.0 (and pylint 1.8) we're incorrectly getting an error that a queryset is not iterable:

```
E: 49,29: Non-iterable value MyModel.objects.filter(foo=bar) is used in an iterating context (not-an-iterable)
```

`MyModel` in this case is a subclass of a model that has `objects` set to a subclass of of the usual manager (in particular, it uses `model_utils.managers.InheritanceManager`)

Similar uses where the standard manager is used does not cause this error, and we did not get this error with previous versions of pylint and pylint-django.

atodorov:
I don't have experience with django-model-utils and I wasn't able to figure out why it produces this error. My search boiled down to `IterableChecker._check_iterable` inside pylint.  After executing `infered = safe_infered(node)` I get:

1) None when the code under evaluation is a standard Django objects manager
2) Const.NoneType(value=None) when inspecting the .all() of an InheritanceManager

Another possibility is to add a transformation similar to `pylint_django/transforms/transforms/django_db_models.py` which sets all of the same attributes of InheritanceManager (and friends).

In fact I think I may go ahead with the transformation option instead of this one. @carlio can I inherit the new transformation class from the above one or do I need to copy everything ?

### DIFF
--- a/pylint_django/augmentations/__init__.py
+++ b/pylint_django/augmentations/__init__.py
@@ -11,7 +11,7 @@ from pylint.checkers.classes import ClassChecker
 from pylint.checkers.newstyle import NewStyleConflictChecker
 from pylint.checkers.variables import VariablesChecker
 from pylint.__pkginfo__ import numversion as PYLINT_VERSION
-from pylint.checkers.typecheck import TypeChecker
+from pylint.checkers.typecheck import IterableChecker, TypeChecker
 from pylint.checkers.variables import ScopeConsumer
 
 from pylint_plugin_utils import augment_visit, suppress_message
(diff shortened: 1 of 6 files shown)


atodorov:
@carlio I'm not quire sure how to use the `_attribute_is_magic` function. Can you take a look? 

I'd prefer to actually be able to check the type of the node being one of several known managers instead of blindly disabling the pylint errors when the attribute is named `objects`. 

Note: this same augmentation function will also fix #144. Also notice that we have the not-an-iterable error reported when the objects manager is defined as class attribute regardless of what the actual manager class is.

PR #145, opened 2018-04-04
Fix explicit unicode check with python_2_unicode_compatible base models

federicobond:
This fixes the case where a `__unicode__` method is defined in a base class via the python_2_unicode_compatible decorator.

I'm a bit lost on how to add a test for this. I would appreciate any help.

### DIFF
--- a/pylint_django/checkers/models.py
+++ b/pylint_django/checkers/models.py
@@ -48,6 +48,27 @@ def _is_meta_with_abstract(node):
     return False
 
 
+def _has_python_2_unicode_compatible_decorator(node):
+    if node.decorators is None:
+        return False
+
+    for decorator in node.decorators.nodes:
+        if getattr(decorator, 'name', None) == 'python_2_unicode_compatible':
+            return True
+
+    return False
+
+
+def _is_unicode_or_str_in_python_2_compatibility(method):
+    if method.name == '__unicode__':
+        return True
+
+    if method.name == '__str__' and _has_python_2_unicode_compatible_decorator(method.parent):
+        return True
+
+    return False
+
+
 class ModelChecker(BaseChecker):
     """Django model checker."""
     __implements__ = IAstroidChecker
@@ -102,12 +123,10 @@ class ModelChecker(BaseChecker):
                 return
 
         # if we get here, then we have no __unicode__ method directly on the class itself
-        if PY3:
-            return
 
         # a different warning is emitted if a parent declares __unicode__
         for method in node.methods():
-            if method.name == '__unicode__':
+            if method.parent != node and _is_unicode_or_str_in_python_2_compatibility(method):
                 # this happens if a parent declares the unicode method but
                 # this node does not
                 self.add_message('W%s03' % BASE_ID, args=node.name, node=node)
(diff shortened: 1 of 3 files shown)


pylint_django/tests/input/func_model_no_explicit_unicode_str_compat.py
```diff
@@ -0,0 +1,19 @@
...
+from django.utils.encoding import python_2_unicode_compatible
+
+
+@python_2_unicode_compatible
+class BaseModel(models.Model):
+    def __str__(self):
+        return 'Foo'
+
+
+class SomeModel(BaseModel):
```
atodorov: Add this class to `func_noerror_unicode_py2_compatible.py` which already includes the rest.

For the record test names usually start with `func_noerror`.

pylint_django/checkers/models.py
```diff
@@ -107,7 +124,7 @@ def _visit_classdef(self, node):
 
         # a different warning is emitted if a parent declares __unicode__
         for method in node.methods():
-            if method.name == '__unicode__':
+            if method.parent != node and _is_unicode_or_str_in_python_2_compatibility(method):
```
atodorov: Isn't `method.parent != node` always True in this context ? This is what the comments suggest?

Also there's a striking similarity between the decorators loop used in your function and the decorators loop used just below this block of code (not shown here). 

Can you explain what you are trying to do because I'm not well familiar with this particular part of the code?

federicobond: > Isn't method.parent != node always True in this context ? This is what the comments suggest?

Not necessarily. The `methods()` generator contains both own and inherited methods. It works for the previous use case because it already asserted that there was no `__unicode__` method in the current class, so if it found one it had to be in one of the ancestors.

> Also there's a striking similarity between the decorators loop used in your function and the decorators loop used just below this block of code (not shown here).

Yes, they are handling two issues derived from the use of the same decorator. I should extract that iteration out. The original code checks for the existence of a decorator defined in the current class. If the decorator exists, it avoids emitting `model-missing-unicode`.

This fix handles the case where the `__unicode__` method is defined not explicitly in an ancestor class but using the decorator. Previously, it wrongly emitted `model-missing-unicode`. Now it emits `model-no-explicit-unicode` because the `__unicode__` method is defined by the decorator in the ancestor class.

pylint_django/tests/input/func_model_no_explicit_unicode_str_compat.py
federicobond: This code should actually emit `model-no-explicit-unicode`. How do I check for that?

atodorov: See *.txt files in the test suite for clues. You have to annotate the lines in the Python source with the expected pylint error, e.g. `# [model-no-explicit-unicode]` and then inside the txt file you describe the error in more details which are `error-key:line_number:class name or empty:error string from pylint (1st line only)`

federicobond:
@atodorov now it appears to build correctly, but one of the build matrix items fails due to missing long_description in setup.py.

atodorov:
@federicobond I've fixed the problem with long_description, the latest Travis build should be fine. Thanks for your PR.

PR #156, opened 2018-04-15
Make get_or_create func placeholder return (model, bool) tuple

femdom:
Otherwise, code like 

```
Model.objects.get_or_create()[0] 
```

raises pylint error.

### DIFF
--- a/pylint_django/transforms/transforms/django_db_models.py
+++ b/pylint_django/transforms/transforms/django_db_models.py
@@ -1,5 +1,11 @@
+import random
+
 from django.core.exceptions import MultipleObjectsReturned \
     as MultipleObjectsReturnedException
+
+
+def __noop_get_or_create(self, *args, **kwargs):
+    return (Model(), bool(random.randint(0, 1)))
 
 
 def __noop(self, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


pylint_django/transforms/transforms/django_db_models.py
```diff
@@ -1,7 +1,13 @@
+import random
+
 from django.core.exceptions import MultipleObjectsReturned \
     as MultipleObjectsReturnedException
 
 
+def __noop_get_or_create(self, *args, **kwargs):
+    return (Model(), bool(random.randint(0, 1)))
```
atodorov: Why use `random` ? Why not return just `True` or `False` ?

atodorov:
@femdom can you post the original error being produced ? For some reason I was not able to reproduce. What pylint_django version do you have ?

pylint_django/transforms/transforms/django_db_models.py
femdom: This implementation is closer to the original one, which may return either `True` or `False` and that makes it more error-proof if in the future pylint gets smart enough to run checks based on exact return values of functions.

atodorov: Fair enough. I still need a reproducer/test though.

daniil-konovalenko:
@atodorov I've faced the same issue:
```
managers_group, _ = Group.objects.get_or_create(
            name='Managers',
 )
```
produces
```
E: Attempting to unpack a non-sequence defined at line 7 of  (unpacking-non-sequence)
```
I use pylint v1.9.1 and pylint-django v0.11.1

PR #158, opened 2018-04-17
New JsonResponse checker

atodorov:
to validate common anti-patterns with http responses returning JSON data.

If there are no objections I will merge this tomorrow morning and release a new version.


carlio:
Ok, already merged but I did check and yes, :+1: from me looks good!

Issue #168, opened 2018-06-26
Proposal: rule to suggest using fields instead of exclude in ModelForm


federicobond:
Using `exclude` in ModelForm.Meta is an antipattern and can easily become a security issue if one is not careful. ModelForms should always define their fields explicitly.

I can contribute a patch if you are interested in this. Also, I'm not sure what to name this rule. `modelform-uses-exclude`?

atodorov:
Please send a PR so I can review.

Can you post more info about Meta.exclude and why it could be a security issue? I'm just curious. 

As for the name, I don't have any particular preferences.

federicobond:
Say you add a BooleanField to a model that controls whether users are allowed to execute a privileged action on a certain object, and default it to False. If you forget to explicitly exclude this field from a form visible to the user, it will now show up as another input ready to be modified and submitted.

If the form data is submitted via AJAX or if the form is rendered field by field, you may not even notice that the new field is now editable.

atodorov:
implemented in #169.

PR #169, opened 2018-06-27
Add modelform-uses-exclude check

federicobond:
Refs #168. The message wording can be improved, but I wanted to get the ball rolling.

### DIFF
--- a/pylint_django/checkers/forms.py
+++ b/pylint_django/checkers/forms.py
@@ -1,0 +1,48 @@
+"""Models."""
+from astroid.nodes import Assign, ClassDef
+
+from pylint.interfaces import IAstroidChecker
+from pylint.checkers.utils import check_messages
+from pylint.checkers import BaseChecker
+
+from pylint_django.__pkginfo__ import BASE_ID
+from pylint_django.utils import node_is_subclass
+
+
+def _get_child_meta(node):
+    for child in node.get_children():
+        if isinstance(child, ClassDef) and child.name == 'Meta':
+            return child
+
+
+class FormChecker(BaseChecker):
+    """Django model checker."""
+    __implements__ = IAstroidChecker
+
+    name = 'django-form-checker'
+    msgs = {
+        'W%d04' % BASE_ID: ("Use explicit fields instead of exclude in ModelForm",
+                            'modelform-uses-exclude',
+                            "Prevents accidentally allowing users to set fields, "
+                            "especially when adding new fields to a Model")
+    }
+
+    @check_messages('modelform-uses-exclude')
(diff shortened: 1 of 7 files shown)


pylint_django/checkers/forms.py
```diff
@@ -0,0 +1,62 @@
...
+
+from pylint_django.__pkginfo__ import BASE_ID
+from pylint_django.utils import node_is_subclass
+
+
+MESSAGES = {
+    'W%d04' % BASE_ID: ("Use explicit fields instead of exclude in ModelForm",
+                        'modelform-uses-exclude',
+                        "Replace exclude with fields attribute"),
+}
```
atodorov: This dictionary is used only inside the checker class so move it there. No need to expose extra objects outside.

pylint_django/checkers/forms.py
```diff
@@ -0,0 +1,62 @@
...
+    """Django model checker."""
+    __implements__ = IAstroidChecker
+
+    name = 'django-form-checker'
+    msgs = MESSAGES
+
+    # NOTE: there's a bug in pylint's backwards compatability logic after changing
+    # visit method names from visit_class to visit_classdef, see
+    # https://bitbucket.org/logilab/pylint/issues/711/new-node-visit-methods-not-backwards
+    if PYLINT_VERSION < (1, 5):
```
atodorov: setup.py requires 'pylint>=1.8.2' so this check is not necessary.

federicobond:
Fixed.

carlio:
Looks good to me after the changes - @atodorov want to double check? Otherwise I think it's good to merge.

pylint_django/checkers/forms.py
```diff
@@ -0,0 +1,49 @@
...
+
+from pylint_django.__pkginfo__ import BASE_ID
+from pylint_django.utils import node_is_subclass
+
+
+def _get_child_meta(node):
+    for child in node.get_children():
+        if isinstance(child, ClassDef) and child.name == 'Meta':
+            return child
+    return None
```
atodorov: You can remove this line b/c this is the default in Python. Indeed the unreleased pylint 2.0 has a check for this and will complain.

federicobond: Yes, I added it because my installed version of Pylint marked it. I'll take it away.

carlio: If pylint complains differently in different versions... well now you know why this PR is fixing the hacks I made :-)  see also https://github.com/PyCQA/pylint-plugin-utils/blob/master/pylint_plugin_utils/__init__.py#L109 ...

PR #170, opened 2018-06-29
Remove legacy check from ModelChecker

federicobond:
The minimum Pylint version is now 1.8.2 as mentioned [here](https://github.com/PyCQA/pylint-django/pull/169#discussion_r198727205), so this check is no longer necessary.


carlio:
Passes tests on Travis so will merge - Thanks @federicobond I was actually about to do this myself too :-)

federicobond:
Great! I'll send another PR with the changes to augmentations.

PR #172, opened 2018-06-29
Upgrading pylint-plugin-utils version

carlio:
I created a new version of pylint-plugin-utils. There are no functional changes - you can see the commits yourself.

Differences now which might impact pylint-django:
  1) Python2 is dropped (which is already true for pylint and pylint-django anyway)
  2) astroid is no longer a direct dependency (which pylint pulls in anyway).

I don't think there will be any problems as the changes are minor.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,7 +17,7 @@ setup(
     packages=find_packages(),
     include_package_data=True,
     install_requires=[
-        'pylint-plugin-utils>=0.2.1',
+        'pylint-plugin-utils>=0.3',
         'pylint>=1.8.2',
     ],
     extras_require={


carlio:
Hmm.

This has a knock on effect. While setup.py does not include python2, the tests do.

The problem is that installing latest pylint-django with this version of pylint-plugin-utils will now cause lots of build failures for people still using python2.

And that means lots of issues being filed here.

So I think the best thing is to revert the pylint-plugin-utils change and keep 'works ok for python2' for now...

atodorov:
Well changelog says this:

```
Version 0.8.0 (20 Jan 2018)

    This is the last version to support Python 2.
```


I am all in favor of dropping Python 2 entirely however I don't understand your concerns.

Looking at Travis CI job for Python 2 (and older Django) this executes just fine. There are some failing tests but the same tests were also failing before your changes.

My vote goes for merging this PR.

carlio:
@atodorov my concerns are that breaking py2 support results in new issues, that's all! I have learned with this projected (and others) that often you get blamed for build failures that are the fault of the dev/test/CI environment not being compatibile. So basically I just thought this change might cause work/hassle for this repo.

I agree in general, I think the PR is a good one. Just wanted to let you know about the possible consequences in terms of time and effort :-)

atodorov:
> @atodorov my concerns are that breaking py2 support results in new issues, that's all!

It's not like you made functional changes that will prevent PY2 from working. All I can see is that we no longer test pylint-plugin-utils under Python 2 which is fine. I am merging this PR.

Issue #165, opened 2018-06-13
QuerySet is inferred as list 


matsui-goga:
If you define the `objects` attribute of a model as `models.Manager()`, you get error
```
E:  5, 3: Instance of 'list' has no 'exists' member (no-member)
```
on
```python
q = AModel.objects.filter(...)
if q.exists():
   ...
```

I'm using pylint 1.9.1, pylint-django 0.11.1.

mahmoud:
Is this not a Manager getting inferred as a list? In any case, yes, I'm seeing this a lot:

```
Instance of 'list' has no 'prefetch_related' member
Instance of 'list' has no 'exclude' member
Instance of 'list' has no 'filter' member
Instance of 'list' has no 'aggregate' member
```

And myriad others. Will try and take a look, see how this can be fixed. Also, on the same versions as above.

atodorov:
```
q = AModel.objects.filter(...)
if q.exists():
```

The problem comes from the second line. In this example `q` is inferred as list and then you get failures.

pylint-django uses transformations which try to teach it what type an object/attribute is and what other attributes there are.

For this to be resolved effectively we need to teach pylint-django about all Manager methods that return a QuerySet.

@carlio ^^^ not sure what I'm getting myself into with this proposition. Ideas ?

carlio:
@atodorov well pylint-django is already a huge shim and things like this exist: https://github.com/PyCQA/pylint-django/blob/master/pylint_django/augmentations/__init__.py#L37 and also https://github.com/PyCQA/pylint-django/blob/master/pylint_django/transforms/fields.py#L31

The whole thing is about second-guessing Django metaprogramming to fit into pylint's expectation of what things are.

So what you're getting yourself into is: adding more special cases to lots of special cases because that's really all that can be done!

mahmoud:
Just a heads up that I'm co-working with @ngtvspc on #175 as a fix for this, and invite anyone with expertise/advice to comment there.

ajl7:
The problem seems to be caused by a transform in pylint-django that substitutes a Manager by something that returns a list for all the methods that return QuerySets: all(), filter(), etc. This doesn't allow for the same methods on the QuerySet itself, so it doesn't recognise chaining, e.g. MyModel.objects.filter(...).select_related(...).order_by(...) 
[...]

PR #175, opened 2018-07-10
WIP: QuerySet Inference

ngtvspc:
This is a rough implementation of a fix for #165.

I believe the issue is here: https://github.com/PyCQA/pylint-django/blob/master/pylint_django/transforms/transforms/django_db_models.py#L43, which causes pylint to infer that the return type of `objects.filter()` is a `list` instead of a `QuerySet`

I can see a couple of ways forward, some of which are called out in the linked issue:
1. Add another stubbed out `QuerySet` similar to the defined `Manager` in `django_db_models.py`
2. Help pylint infer the return types of Manager methods
3. Revisit the problems with import Django itself and allowing pylint to infer the actual classes themselves

I've opted to try to implement 2, however I can see it's missing two key things:
1. Specific detection and restriction of the type the `QuerySet` returning functions are called on. (namely it needs to be only QuerySets and Managers)
2. Since it infers the type as the generic `QuerySet`, it is still missing the custom methods defined on the specific `ModelQuerySet`

If someone could point me in the right direction I would appreciate it

### DIFF
--- a/pylint_django/tests/input/func_noerror_managers_return_querysets.py
+++ b/pylint_django/tests/input/func_noerror_managers_return_querysets.py
@@ -1,0 +1,50 @@
+"""
+Checks that Pylint does not complain about Manager methods returning lists
+when they in fact return QuerySets
+"""
+
+from django.db import models
+
+class SomeModelQuerySet(models.QuerySet):
+    """ A missing docstring """
+    @staticmethod
+    def public_method():
+        """ A missing docstring """
+        return 1
+
+
+class SomeModelManager(models.Manager):
+    """A missing docstring"""
+
+    @staticmethod
+    def public_method():
+        """ A missing docstring """
+        return 1
+
+    @staticmethod
+    def another_public_method():
+        """ A missing docstring """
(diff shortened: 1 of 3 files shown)


atodorov:
> I believe the issue is here: https://github.com/PyCQA/pylint-django/blob/master/pylint_django/transforms/transforms/django_db_models.py#L43, which causes pylint to infer that the return type of objects.filter() is a list instead of a QuerySet

this is correct. 

I'm not sure I like the approach where we try to teach pylint_django how to infer things. In fact I'm starting to dislike this b/c we have to take into account every possible edge case.  I have not looked in details at your source code though. Will do so later.

Have you tried configuring Django instead ? From what I can tell the user needs to configure DJANGO_SETTINGS_MODULE in their environment and assuming Python can import this we should be good to go.

ngtvspc:
I haven't tried the approach yet of configuring Django, I agree that would be preferable because I can already see edge cases this doesn't handle because of inheritance and custom `QuerySet`. I'll do a little digging to figure out what that would take

PR #178, opened 2018-07-15
Migrate to pylint 2.0

atodorov:
Initial work to support pylint 2.0. However many tests started failing. 

Tips are welcome, I will look at the tests in the next couple of days.

@carlio what do you think about requiring pylint >= 2.0 from now on and not dealing with backwards compatibility for newer versions ?

### DIFF
--- a/pylint_django/__init__.py
+++ b/pylint_django/__init__.py
@@ -2,12 +2,11 @@
 from __future__ import absolute_import
 
 import sys
-import warnings
 
 from pylint_django import plugin
 
 if sys.version_info < (3, ):
-    warnings.warn("Version 0.8.0 is the last to support Python 2. "
-                  "Please migrate to Python 3!", DeprecationWarning)
+    raise DeprecationWarning("Version 0.11.1 was the last to support Python 2. "
+                             "Please migrate to Python 3!")
 
 register = plugin.register  # pylint: disable=invalid-name
--- a/pylint_django/transforms/foreignkey.py
+++ b/pylint_django/transforms/foreignkey.py
@@ -24,7 +24,7 @@ def infer_key_classes(node, context=None):
     for arg in node.args:
         # typically the class of the foreign key will
         # be the first argument, so we'll go from left to right
-        if isinstance(arg, (nodes.Name, nodes.Getattr)):
+        if isinstance(arg, (nodes.Name, nodes.Attribute)):
             try:
                 key_cls = None
                 for inferred in arg.infer(context=context):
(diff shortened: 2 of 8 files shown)


pylint_django/__init__.py
```diff
@@ -2,12 +2,11 @@
...
 
 import sys
-import warnings
 
 from pylint_django import plugin
 
 if sys.version_info < (3, ):
-    warnings.warn("Version 0.8.0 is the last to support Python 2. "
-                  "Please migrate to Python 3!", DeprecationWarning)
+    raise DepracationWarning("Version 0.11.1 was the last to support Python 2. "
```
federicobond: typo: `DeprecationWarning`

atodorov:
Looks like we are in business now. I see some failures which look like the linter was not ignoring scenarios it should have ignored.

carlio:
@atodorov I think that an aggressive requirements update is a good plan because that is how Django and pylint and astroid are doing things; pypi still has old pylint-django, so I'd agree, make it pylint>=2.0 and perhaps just an extra line or two in the README to mention that change.

If people don't pin even test dependencies they can't reasonably expect their builds to work all the time, and I prefer making it their problem rather than ours by adding too much backwards compatabilty overhead. As I said elsewhere - if it's easy, why not, but let's not try too hard.

There is a PR for bumping pylint-plugin-utils to 0.4 which will include pylint2 compatability - that's here: https://github.com/PyCQA/pylint-django/pull/180

pylint_django/transforms/foreignkey.py
```diff
@@ -24,7 +24,7 @@ def infer_key_classes(node, context=None):
     for arg in node.args:
         # typically the class of the foreign key will
         # be the first argument, so we'll go from left to right
-        if isinstance(arg, (nodes.Name, nodes.Getattr)):
+        if isinstance(arg, (nodes.Name, nodes.Attributes)):
```
federicobond: typo: `nodes.Attribute`

federicobond:
@atodorov I have a couple more fixes that build on top of this. How should we go about doing it? `master` is red anyway, so I don't see much trouble merging this to keep momentum going.

atodorov:
@federicobond can you at least share a branch that we can take a look at? 

I've managed to narrow things down to only 5 failing tests and I don't mind merging to master to allow you or others to send in pull requests.

federicobond:
I have 5 too, my fixes matched your latest two commits.

atodorov:
Merged so we can allow others to work on this if they like. I will continue investigating the failing tests in the upcoming days.

PR #179, opened 2018-07-16
Fix build with Python 3.7

federicobond:


federicobond:
@atodorov I think the config is good now. It fails with the same error as `master`.

.travis.yml
```diff
@@ -18,13 +19,15 @@ matrix:
...
     - { python: 2.7, env: DJANGO=2.0 }
+    - { python: "3.7-dev", env: DJANGO=1.11 }
   include:
     - { stage: django_not_installed, python: 3.6, env: TOXENV=django_not_installed }
     - { stage: django_is_installed, python: 3.6, env: TOXENV=django_is_installed }
     - { stage: test, python: 3.6, env: TOXENV=flake8 }
     - { stage: test, python: 3.6, env: TOXENV=pylint }
     - { stage: test, python: 3.6, env: TOXENV=readme }
     - { stage: build_and_package_sanity, python: 3.6, env: SANITY_CHECK=1 }
+    - { stage: test, python: 3.7-dev, env: DJANGO=2.0 }
```
atodorov: You don't need this I think. IMO this is the cause of 2 py3.7 jobs being created. See jobs .9 and .12 in the Travis matrix.

federicobond:
@atodorov done!

PR #183, opened 2018-07-23
Internal updates caused by pylint/astroid 2.0. Fixes #182.

### LINKED ISSUE #182: HIGH: Failing tests with pylint 2.0 and astroid 2.0
This is to comment on the still failing tests in master. This needs to be fixed before we can release a new version.

CC @carlio @federicobond @PCManticore

atodorov:
Notes:

1) The first commit will clash with the latest commit on top of master. I rewrote the commit for a shorter implementation and then realized we had already merged the earlier implementation.

2) Take a look at the new way we apply transformations. I'm not quite certain this is what we need to be doing here but it fixed couple of failures, namely:
https://travis-ci.org/PyCQA/pylint-django/builds/405211436
`func_noerror_model_fields` and `func_noerror_ugettext_lazy_format`


atodorov:
Excluding the conflict I now have only 1 failing test. If I apply the following diff:

```
--- a/pylint_django/tests/test_func.py
+++ b/pylint_django/tests/test_func.py
@@ -43,6 +43,8 @@ def get_tests(input_dir='input', sort=False):
 
     suite = []
     for fname in os.listdir(input_dir):
+        if fname.find('func_noerror_foreign') == -1:
+            continue
         if fname != '__init__.py' and fname.endswith('.py'):
             suite.append(test_functional.FunctionalTestFile(input_dir, fname))
```

then I get only a subset of the tests executed (including the failing one) and this time it passes. This looks very strange.

atodorov:
@PCManticore any idea why I get varying results depending on how I execute one of the tests ?

pylint_django/transforms/__init__.py
```diff
@@ -2,53 +2,48 @@
...
+        with open(fake_module_path) as modulefile:
+            fake_module = modulefile.read()
+
+        return astroid.builder.AstroidBuilder(astroid.MANAGER).string_build(fake_module)
+
+    astroid.register_module_extender(astroid.MANAGER, package_name, fake_module_builder)
+
+# TODO: no sure what to do with commented out code! It looks like
+# we don't need these transforms anymore
+
```
carlio: I'd suggest either leave them in or just a comment with a commit hash and an explanation of 'we removed some transforms' in case new issues get opened wanting them back.

carlio:
Code seems to make sense to me but we're still getting build failures.

atodorov:
@carlio sorry for the noise. Finally figured out what was wrong with the test. See the last commit. 

I am up for a release tomorrow, any preferences for teh version number ? Should we also do 2.0 or keep it incremental (0.12 then).

carlio:
@atodorov Yes I think it's time to have a better version scheme. It just started as I used '0' to mean 'not quite sure this is done yet' and never changed it. I suppose it might as well follow pylint more closely but at least it could get a major version number now I think.

carlio:
@atodorov and if that number  changes, could probably remove the [beta classifier](https://github.com/PyCQA/pylint-django/blob/master/setup.py#L28) at the same time.

PR #193, opened 2018-08-25
Suppress false-positive no-self-argument in factory.post_generation

### LINKED ISSUE #190: False-positive no-self-argument in factory.post_generation method
In factory_boy, post_generation methods defined via the `factory.post_generation` decorator receive the model object, not self (as in the factory) as first argument. We should silence these errors.

Reference: https://factoryboy.readthedocs.io/en/latest/reference.html?highlight=post_generation#extracting-parameters (note the first argument is called `obj`, not `self`)

federicobond:
Fixes #190

### DIFF
--- a/pylint_django/augmentations/__init__.py
+++ b/pylint_django/augmentations/__init__.py
@@ -447,6 +447,23 @@ def is_model_factory(node):
     return False
 
 
+def is_factory_post_generation_method(node):
+    if not node.decorators:
+        return False
+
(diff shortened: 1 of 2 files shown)


federicobond:
@atodorov thank you for cutting a release so soon! I wasn't expecting it.

carlio:
@federicobond that's how we roll in pylint-django-land :-)

PR #207, opened 2018-11-29
Cleaning

carlio:
@atodorov placholder PR to clean up the old compat code that is no longer necessary

### DIFF
--- a/pylint_django/compat.py
+++ b/pylint_django/compat.py
@@ -22,10 +22,3 @@ except ImportError:
     except ImportError:
         from astroid.util import Uninferable
 
-try:
-    django = __import__("django")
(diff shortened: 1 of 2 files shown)


pylint_django/compat.py
```diff
@@ -28,16 +29,3 @@
     # if not available, will be handled by the django_installed checker
     django_version = (1, 5)
```
atodorov: FIY we're Python 3 only ATM so no need to worry about older Django versions, right ?

carlio: The hardcoded `1.5` was really just to try to avoid complete failures due to not being able to import django at all. I will look through this but I think this method can go completely too.

Issue #120, opened 2018-01-23
TypeError: 'NamesConsumer' object does not support indexing


simobasso:
how to reproduce:

freeze
```
astroid==1.6.1
Django==1.11
isort==4.2.15
lazy-object-proxy==1.3.1
mccabe==0.6.1
pylint==1.8.2
pylint-django==0.8.0
pylint-plugin-utils==0.2.6
pytz==2017.3
six==1.11.0
wrapt==1.10.11
```

R0201.py
```python
class Foo(object):
    def bar(self, baz):
        return 33
```

command
```
pylint --load-plugins=pylint_django  R0201.py
```

output
```
No config file found, using default configuration
************* Module R0201
C:  1, 0: Module name "R0201" doesn't conform to snake_case naming style (invalid-name)
C:  1, 0: Missing module docstring (missing-docstring)
C:  1, 0: Missing class docstring (missing-docstring)
C:  2, 4: Black listed name "bar" (blacklisted-name)
C:  2, 4: Black listed name "baz" (blacklisted-name)
C:  2, 4: Missing method docstring (missing-docstring)
W:  2,18: Unused argument 'baz' (unused-argument)
R:  2, 4: Method could be a function (no-self-use)
R:  1, 0: Too few public methods (1/2) (too-few-public-methods)
Traceback (most recent call last):
  File "~/.pyenv/versions/pylint-django-try/bin/pylint", line 11, in <module>
    sys.exit(run_pylint())
  File "~/.pyenv/versions/3.6.3/envs/pylint-django-try/lib/python3.6/site-packages/pylint/__init__.py", line 16, in run_pylint
    Run(sys.argv[1:])
  File "~/.pyenv/versions/3.6.3/envs/pylint-django-try/lib/python3.6/site-packages/pylint/lint.py", line 1347, in __init__
    linter.check(args)
  File "~/.pyenv/versions/3.6.3/envs/pylint-django-try/lib/python3.6/site-packages/pylint/lint.py", line 768, in check
    self._do_check(files_or_modules)
  File "~/.pyenv/versions/3.6.3/envs/pylint-django-try/lib/python3.6/site-packages/pylint/lint.py", line 901, in _do_check
    self.check_astroid_module(ast_node, walker, rawcheckers, tokencheckers)
  File "~/.pyenv/versions/3.6.3/envs/pylint-django-try/lib/python3.6/site-packages/pylint/lint.py", line 980, in check_astroid_module
    walker.walk(ast_node)
[...]

PR #209, opened 2018-12-14
Backport #120 to 0.8 series

atodorov:
@nedbat, @carlio I am planning to merge this one and release a 0.8.1 version with the added note that it is not tested and will not be further supported.


carlio:
@atodorov no complaints from me, seems the most pragmatic solution.

PR #214, opened 2018-12-20
Add support for new load_configuration hook of pylint

### LINKED ISSUE #181: Custom good names
After change in pull request https://github.com/PyCQA/pylint-django/pull/109 seems to be broken to ability to set own good names. Example pylintrc:

```
[MASTER]
load-plugins=pylint_django

[BASIC]
good-names=_,dt,e,f,default_app_config,logger
```

In version v0.7.2 it was working properly but starting version 0.8.0 it can be one or other. Probably previous version was merging two lists of good names which is not the case now.

matusvalo:
This PR adds new pylint hook `load_configuration()`. This hook serves for adjusting configuration of linter: http://pylint.pycqa.org/en/latest/how_tos/plugins.html

Closes #181

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -9,3 +9,4 @@
 * [atodorov](https://github.com/atodorov)
 * [bittner](https://github.com/bittner)
 * [federicobond](https://github.com/federicobond)
+* [matusvalo](https://github.com/matusvalo)
--- a/pylint_django/checkers/db_performance.py
+++ b/pylint_django/checkers/db_performance.py
@@ -13,6 +13,7 @@ from pylint import interfaces
 from pylint import checkers
 from pylint.checkers import utils
 from pylint_django.__pkginfo__ import BASE_ID
+from pylint_django import compat
 
 
 def _is_addfield_with_default(call):
@@ -113,12 +114,16 @@ class NewDbFieldWithDefaultChecker(checkers.BaseChecker):
                     self.add_message('new-db-field-with-default', args=module.name, node=node)
 
 
-def register(linter):
-    """Required method to auto register this checker."""
+def load_configuration(linter):
     # don't blacklist migrations for this checker
     new_black_list = list(linter.config.black_list)
     if 'migrations' in new_black_list:
         new_black_list.remove('migrations')
     linter.config.black_list = new_black_list
 
+
+def register(linter):
+    """Required method to auto register this checker."""
     linter.register_checker(NewDbFieldWithDefaultChecker(linter))
+    if not compat.LOAD_CONFIGURATION_SUPPORTED:
+        load_configuration(linter)
--- a/pylint_django/compat.py
+++ b/pylint_django/compat.py
@@ -22,3 +22,7 @@ except ImportError:
     except ImportError:
         from astroid.util import Uninferable
 
+import pylint
+
+# pylint before version 2.3 does not support load_configuration() hook.
+LOAD_CONFIGURATION_SUPPORTED = pylint.__pkginfo__.numversion >= (2, 3)
--- a/pylint_django/plugin.py
+++ b/pylint_django/plugin.py
@@ -7,13 +7,12 @@ from pylint_django.checkers import register_checkers
 # we want to import the transforms to make sure they get added to the astroid manager,
 # however we don't actually access them directly, so we'll disable the warning
 from pylint_django import transforms  # noqa, pylint: disable=unused-import
+from pylint_django import compat
 
 
-def register(linter):
+def load_configuration(linter):
     """
-    Registering additional checkers.
-
-    However, we will also use it to amend existing checker config.
+    Amend existing checker config.
     """
     name_checker = get_checker(linter, NameChecker)
     name_checker.config.good_names += ('qs', 'urlpatterns', 'register', 'app_name', 'handler500')
@@ -21,6 +20,11 @@ def register(linter):
     # we don't care about South migrations
     linter.config.black_list += ('migrations', 'south_migrations')
 
+
+def register(linter):
+    """
+    Registering additional checkers.
+    """
     # add all of the checkers
     register_checkers(linter)
 
@@ -32,3 +36,6 @@ def register(linter):
         # probably trying to execute pylint_django when Django isn't installed
         # in this case the django-not-installed checker will kick-in
         pass
+
+    if not compat.LOAD_CONFIGURATION_SUPPORTED:
+        load_configuration(linter)


matusvalo:
This PR can be closed **after** releasing new version of pylint with https://github.com/PyCQA/pylint/pull/2644

matusvalo:
@atodorov I have fixed also `db_performance.py` checker. Regarding tests, I don't expect them to be green. I suppose that first dependency to pylint needs to be fixed:
https://github.com/PyCQA/pylint-django/blob/bb0153ba1388e5304d5c902efa0d260883e99d1a/setup.py#L19-L22
Now, the change in pylint is just in master branch, we need to wait to release, and bump dependency of pylint to most recent version. After tests can be fixed. Or is there any other way how to proceed?

carlio:
Perhaps I misunderstand but this would not be backwards compatible with older pylints as `load_configuration` would no longer be called by them? So merging this would make anything not using the newest pylint suddenly start showing errors that were not there before. 

I think that calling `load_configuration` explicitly in `register` would keep both functionalities for now, perhaps with a conditional based on pylint version, and certainly a comment explaining the pylint version that made the change for later compatability cleanups.

matusvalo:
@carlio You are right. This PR needs to bump dependency to pylint version which supports `load_configuration()` hook. For now this hook is just in master branch and will be released sometime in future - this was also mentioned in my first comment:

> This PR can be closed **after** releasing new version of pylint with [PyCQA/pylint#2644](https://github.com/PyCQA/pylint/pull/2644)

I am fine with adjusting this PR with backward compatibility - when older version of pylint is detected, `load_configuration()` will be called from `register()`. But this can be confusing for users since there is open issue https://github.com/PyCQA/pylint-django/issues/181 which will affect users when they have older pylint installed.

carlio:
@matusvalo the [compat.py](https://github.com/PyCQA/pylint-django/blob/master/pylint_django/compat.py) file is exactly for things like this. The problem is the number of configurations for build systems of the type `pylint==X, pylint-django` so there needs to be a way make non-breaking changes more gently than rigid requirements limits on pylint.

@atodorov and I have had this discussion many times, I think he's more on your side but I certainly prefer making a smooth movement given how often a change to pylint-django can break a build. A simple 'if pylint_version < x.y.z: load_configuration()' isn't much and could be in compat.py with a good comment so it can be removed in the future.

We do need to figure out a better way to warn about things like this, perhaps a new emitted warning our regular python `warn` about mild version incompatabilities. Anyway I just want to avoid fixing #181 then ending up with a new issue from users with older pylints but unpinned pylint-django dependencies. That's equally confusing!

matusvalo:
@carlio I am fine with your proposal. I will implement your comments.

atodorov:
> 'if pylint_version < x.y.z: load_configuration()' isn't much and could be in compat.py with a good comment so it can be removed in the future.

+1 for that.

Re #181 - there's no way around that if people are using an older version of pylint. We'll document this when releasing the new version but really if you insist on using an older pylint there's nothing we can do about it.

I personally favor using the latest and greatest versions of dev tools but not everybody does this unfortunately.

carlio:
@matusvalo oh btw don't forget to add the change to the CHANGELOG if you want. No worries it will get done for next release though 👍

pylint_django/__init__.py
```diff
@@ -3,10 +3,17 @@
...
 if sys.version_info < (3, ):
     raise DeprecationWarning("Version 0.11.1 was the last to support Python 2. "
                              "Please migrate to Python 3!")
 
-register = plugin.register  # pylint: disable=invalid-name
+# Older versions of pylint does not support load_configuration() hook. To be
+# backward compatible we call load_configuration() directly from register()
+# hook. Using legacy approach yields issue #181.
+# For more details see issue #181 and PR #214.
+if compat.REGISTER_LEGACY:
```
carlio: wait is this the right way round? `compat.REGISTER_LEGACY` sounds like it should call `plugin.register_legacy`. Might just be a naming issue

matusvalo: Oh stupid mistake... I will fix it. In any case I would like to do some manual tests for this functionality + investigate whether we can print some warning when old pylint is used. Maybe some unittests can be added for this functionality what do you think?

matusvalo:
Btw. I manually tried the PR and:

Current pylint version (2.2.2) works as expected:
```
$ pylint --load-plugins pylint_django foo/foo/
************* Module foo
foo/foo/__init__.py:1:0: C0102: Black listed name "foo" (blacklisted-name)

------------------------------------------------------------------
Your code has been rated at 9.60/10 (previous run: 8.80/10, +0.80)


$ pylint --load-plugins pylint_django --good-names=foo foo/foo/
************* Module foo.urls
foo/foo/urls.py:19:0: C0103: Constant name "urlpatterns" doesn't conform to UPPER_CASE naming style (invalid-name)

------------------------------------------------------------------
Your code has been rated at 9.60/10 (previous run: 9.60/10, +0.00)
```

**BUT** with with pylint from master branch I am getting the following stacktrace:
```
$ pylint --load-plugins pylint_django foo/foo/
  File "/home/matus/environments/pylint36/bin/pylint", line 11, in <module>
    load_entry_point('pylint==2.3.0', 'console_scripts', 'pylint')()
  File "/home/matus/environments/pylint36/lib/python3.6/site-packages/pylint/__init__.py", line 20, in run_pylint
    Run(sys.argv[1:])
  File "/home/matus/environments/pylint36/lib/python3.6/site-packages/pylint/lint.py", line 1537, in __init__
    linter.load_plugin_modules(self._plugins)
  File "/home/matus/environments/pylint36/lib/python3.6/site-packages/pylint/lint.py", line 655, in load_plugin_modules
    module.register(self)
  File "/home/matus/environments/pylint36/lib/python3.6/site-packages/pylint_django/plugin.py", line 30, in register
    apply_augmentations(linter)
  File "/home/matus/environments/pylint36/lib/python3.6/site-packages/pylint_django/augmentations/__init__.py", line 763, in apply_augmentations
    suppress_message(linter, TypeChecker.visit_attribute, 'no-member', is_model_field_display_method)
  File "/home/matus/environments/pylint36/lib/python3.6/site-packages/pylint_plugin_utils/__init__.py", line 125, in suppress_message
    get_message_definition = msgs_store.get_message_definition
AttributeError: 'MessagesStore' object has no attribute 'get_message_definition'
```
And this is not connected to this PR since it occurs also on master branch of pylint-django. I think it is connected to this commit:
https://github.com/PyCQA/pylint/commit/da67a9da682e51844fbc674229ff6619eb9c816a#diff-9f2e05113cbb502c875114aa3287baef

I am not able to verify this PR until this is handled :-(

carlio:
@matusvalo welcome to the world of trying to keep compatability between pylint upstream and users' configurations downstream :-)  This is why I always push to have that `compat.py` even though it is gunge in the codebase it is unfortunately a necessary evil.

matusvalo:
@carlio my opinion is that this is not ideal for users in any case: When user is not using latest version pylint-django he does not know which version of pylint he can use because setup.py specifies only lower bound:
https://github.com/PyCQA/pylint-django/blob/bb0153ba1388e5304d5c902efa0d260883e99d1a/setup.py#L21

E.g. for instance if pylint 2.3 is released and user uses combination of pylint 2.3 and pylint-django 2.0.5 he will get exception. Don't get me wrong. I am not saying that this should be fixed on pylint-django side - this would be very difficult to fix if not impossible. (I think better is to have stable API for plugins on pylint side).

matusvalo:
@carlio @atodorov I am proposing the following implementation of warnings when "old" version of pylint is detected:
```diff
$ git diff
diff --git a/pylint_django/compat.py b/pylint_django/compat.py
index a461548..0393082 100644
--- a/pylint_django/compat.py
+++ b/pylint_django/compat.py
@@ -2,6 +2,10 @@
 # pylint: skip-file
 # no sane linter can figure out the hackiness in this compatability layer...

+import warnings
+
+warnings.simplefilter('always', DeprecationWarning)
+
 try:
     from astroid.nodes import ClassDef, FunctionDef, ImportFrom, AssignName, Attribute
 except ImportError:
@@ -26,3 +30,11 @@ import pylint

 # pylint before version 2.3 does not support load_configuration() hook.
 REGISTER_LEGACY = pylint.__pkginfo__.numversion < (2, 3)
+
+if REGISTER_LEGACY:
+    warnings.warn(
+        "Legacy pylint version is detected. Custom configuration is not supported. "
+        "Please upgrade pylint to version >=2.3. For more details see: "
+        "https://github.com/PyCQA/pylint-django/issues/181",
+        DeprecationWarning
+    )
```
For user it will be displayed as follows:
```
$ pylint --load-plugins pylint_django --good-names=foo foo/foo/
/home/d56032/environments/pylint36/lib/python3.6/site-packages/pylint_django-2.0.5-py3.6.egg/pylint_django/compat.py:39: DeprecationWarning: Legacy pylint version is detected. Custom configuration is not supported. Please upgrade pylint to version >=2.3. For more details see: https://github.com/PyCQA/pylint-django/issues/181
  DeprecationWarning
************* Module foo.urls
foo/foo/urls.py:19:0: C0103: Constant name "urlpatterns" doesn't conform to UPPER_CASE naming style (invalid-name)

------------------------------------------------------------------
Your code has been rated at 9.60/10 (previous run: 9.60/10, +0.00)
```
Are you fine with it? If yes I will add it to this PR.

pylint_django/__init__.py
atodorov: -1 on this syntax for the following reasons:

* `__init__.py` should be as simple as possible
* `db_performance.py::register()` calls `load_configuration()` based on this condition. Why should `plugin.py::register()` be different ? 
* Also consider maintainability and other contributors down the road: While `register()` is a function provided by pylint's API interface `register_legacy()` is not. Let's save ourselves the troubles of figuring out what that was or even missing it entirely sometime down the road. (these things happen, I can show you a very recent example where I screwed up a PR and then an entire team of maintainers failed to see my mistake, pushed new version upstream and screwed up even more. We also had a similar less spectacular scenario in pylint_django recently.).

To be clear leave `__init__.py` alone and put the if condition inside `plugin.register()` the same way you do in `db_performance.register()`.

atodorov: @carlio, @matusvalo - looking at the latest test results I see all PASS. See job 524.8 with env DJANGO=master. This is installing pylint from their master branch and the console log shows pylint version 2.3.0.

IMO we're OK on the testing side here.

atodorov:
@matusvalo about your proposal for warnings:

It is out of scope for this PR IMO. I am open to discussing this in another PR or issue but please don't add it here. I don't mind the warning but honestly for people who need to use older versions for whatever reason this warning doesn't bring them any good. 

@carlio - the above mentioned traceback may be a future breakage waiting to happen in pylint-plugin-utils. I'll try adding the same test config with pylint@master that we have here.

matusvalo:
@atodorov all commits are squashed now.

atodorov:
Merged. I suggest we release a new version when pylint 2.3.0 comes out.

PR #224, opened 2019-04-11
Support ForeignKey relations with `to` keyword. Fixes #223 

### LINKED ISSUE #223: pylint-django doesn't recognise ForeignKey models when using the `to` kwarg
In addition to not recognising the model when it's supplied as a string, the following does not work

`related = models.ForeignKey(to=Other, on_delete=models.SET_NULL)`

the linter works as expected with

`related = models.ForeignKey(Other, on_delete=models.SET_NULL)`

Is it possible to allow kwarg usage? I prefer being explicit so that it's easy to find all references to a model.

fadedDexofan:
Added support for `to` keyword FK relations. See https://github.com/PyCQA/pylint-django/issues/223

### DIFF
--- a/pylint_django/tests/input/func_noerror_foreignkeys.py
+++ b/pylint_django/tests/input/func_noerror_foreignkeys.py
@@ -7,6 +7,10 @@ from django.db import models
 from django.db.models import ForeignKey, OneToOneField
 
 
+class Genre(models.Model):
+    name = models.CharField(max_length=100)
+
+
 class Author(models.Model):
     author_name = models.CharField(max_length=100)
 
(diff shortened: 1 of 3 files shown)


pylint_django/tests/input/func_noerror_foreignkeys.py
```diff
@@ -17,8 +21,9 @@ class ISBN(models.Model):
 
 class Book(models.Model):
     book_name = models.CharField(max_length=100)
-    author = models.ForeignKey('Author', on_delete=models.CASCADE)
-    isbn = models.OneToOneField(ISBN, on_delete=models.CASCADE)
+    author = models.ForeignKey(to='Author', on_delete=models.CASCADE)
+    isbn = models.OneToOneField(to=ISBN, on_delete=models.CASCADE)
+    genre = models.ForeignKey(Genre, on_delete=models.CASCADE)
```
atodorov: Maybe add a comment that this last FK is missing the `to` keyword on purpose.  I see there are other models in the test suite which are missing the `to` keyword (and that is fine). I just want to make sure we do not accidentally clean up this use case from the test suite and end up breaking things.

Issue #198, opened 2018-09-25
Custom error code for not loaded related model


dmytrokyrychuk:
In my opinion `Instance of 'ForeignKey' has no '..' member` errors that show up on ForeignKeys that referend models with a string path is not an issue with `pylint_django` plugin, but rather with the code that is being checked. It is better to communicate this to the plugin users via an error message tied directly to the ForeignKey, rather than via the plugin documentation.

When a model that is referenced by `'app_name.ModelName'` string is not loaded when `pylint_django` inspect a model class, it should produce a new error (e.g., `E5102: Referenced model 'app_name.ModelName' not loaded.`). This way it will be much clearer to the user what is the issue with the code that they wrote. If the user need the circular dependency there, they would be able to disable the error by adding `# pylint: disable=E5102` to the line where ForeignKey is defined.

It would be great if `no-member` errors related to that ForeignKey were suppressed by `pylint_django`, so that if the user would only need to suppress the original `E5102` error on the line where ForeignKey is defined, instead of on every line where it is referenced. Does anyone know if it is possible to suppress errors like this in pylint?

atodorov:
@carlio again another one related to #188 and #192. I'm really wondering if we should try to load Django before doing anything else in the hope that all of these references and auto-magical attributes will get resolved. Maybe I will give it a try and see how it goes.

carlio:
@atodorov I missed this reference my apologies.

Originally Django was not loaded because any importing of `django.db.models` resulted in settings being loaded and it would fail due to settings not being properly configured.

However that was many Django versions ago and it could be the case that the range of Django versions supported now do not have this problem (https://docs.djangoproject.com/en/2.2/topics/settings/#using-settings-without-setting-django-settings-module) so loading Django might not immediately cause breakage any more.

carlio:
It could however lead to default settings that impact the code people are using pyling-django to check, so to avoid that I think leaving this as pure linting rather than depending on Django would make things easier and less brittle.

atodorov:
With the changes from #229 merged I think the original root cause of this issue has been resolved.

IMO there is no need to add a custom error message anymore and no need trying to load django settings (for now :)).

As @imomaliev suggested in #229:
> close until another issues related to this emerge.

@carlio, @dmytrokyrychuk what do you think ?

carlio:
@atodorov I say close it and if it isn't fixed, we can reopen or create a new issue.

PR #229, opened 2019-04-21
FK as string from models as package fails with "UnboundLocalError: local variable 'key_cls' referenced before assignment"

imomaliev:
Currently ForeignKey inferring fails with `UnboundLocalError: local variable 'key_cls' referenced before assignment` for cases  when models is a python package, the `to` argument is a string that is used in this pattern `app.Model` and also there is some  other `bool` const like `null=True` right after `to`.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,10 +1,10 @@
 Changelog
 =========
 
-Version 2.0.8 (18 April 2019)
+Version 2.0.9 (dev)
 -----------------------------
 
-- Fix `UnboundLocalError: local variable 'key_cls' referenced before assignment`` for FK when using ``to`` as string pointing to model when models is a package
+- Fix ``UnboundLocalError: local variable 'key_cls' referenced before assignment`` for FK when using ``to`` as string pointing to model when models is a package
   
 Version 2.0.8 (18 April 2019)
 -----------------------------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,11 @@ Changelog
 Version 2.0.8 (18 April 2019)
 -----------------------------
 
+- Fix `UnboundLocalError: local variable 'key_cls' referenced before assignment`` for FK when using ``to`` as string pointing to model when models is a package
+  
+Version 2.0.8 (18 April 2019)
+-----------------------------
+
 - Support recursive (self) ForeignKey relations. Fix
   `#208 <https://github.com/PyCQA/pylint-django/issues/208>`_ (Daniil Kharkov)
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,7 +4,7 @@ Changelog
 Version 2.0.9 (dev)
 -----------------------------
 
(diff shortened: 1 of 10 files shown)


CHANGELOG.rst
```diff
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+Version 2.0.9 (dev)
+-----------------------------
+
+- Fix ``UnboundLocalError: local variable 'key_cls' referenced before assignment`` for FK when using ``to`` as string pointing to model when models is a package
```
atodorov: Don't be shy and add your name (or github handle if you prefer) in brackets after this line. We like to show who contributed what.

pylint_django/tests/input/models/author.py
```diff
@@ -0,0 +1,6 @@
+# pylint: disable=missing-docstring,wrong-import-position
```
atodorov: `wrong-import-position` is probably not needed here.

pylint_django/tests/input/models/func_noerror_foreign_key_key_cls_unbound.py
```diff
@@ -0,0 +1,14 @@
+"""
+Checks that Pylint does not complain about ForeignKey pointing to model
+in module of models package
+"""
+# pylint: disable=missing-docstring,wrong-import-position
```
atodorov: `wrong-import-position`

pylint_django/tests/test_func.py
```diff
@@ -56,6 +56,7 @@ def _file_name(test):
 
 
 TESTS = get_tests()
+TESTS.extend(get_tests('input/models'))
```
atodorov: What if you move your `models/__init__.py` changes into `tests/input/__init__.py` and adjust the location of the other files as well? So maybe we'll have something like
```
input/
        __init__.py
       authors.py (or models/authors.py)
      ... tests ...
```

I just don't want to start having multiple sub-directories if not really necessary.

atodorov:
@imomaliev is this by any chance the fix for #198 ? If so please add 'Fix #198' to one of the commit messages and GitHub will close it automatically.

pylint_django/tests/test_func.py
imomaliev: We don't need to load `'input/models'` in tests, but we need `models` package here, because it is required for FK inferring, `infer_key_classes` adds `.models` on  https://github.com/PyCQA/pylint-django/blob/87f9b4cfcaeeb212d2eb8c6889ea1f0760db9936/pylint_django/transforms/foreignkey.py#L91.

imomaliev:
@atodorov made requested changes and moved tests to input folder, `models` package is still required. Regarding #198, looks like it is related and should probably fix it. But what OP is asking for is separate error code for cases when `models` are not loaded in astroid for some reason. I suggest closing until another issues related to this emerge. Should I add it commit message?

prokher:
Just received the same error in `2.0.9`.

carlio:
Also here - #232

zout:
Also posted in #232 but no response yet. But I've narrowed it down to OneToOne relations. On my github there is a project that will trigger the error. Hope it helps in finding out the root cause: https://github.com/zout/django-pylint-issue

PR #236, opened 2019-06-21
Bugfix: Use functools.wrap to preserve leave_module info

mohi7solanki:
Since pylint-plugin-utils's `augment_visit` was using checker_methods `__module__` and `__qualname__` to get the checker [visit_augment](https://github.com/PyCQA/pylint-plugin-utils/blob/43fde2104fb4e7d6aa3a6a71725bca31d44eb936/pylint_plugin_utils/__init__.py#L50) this information was lost after decorating the `leave_module` [wrap](https://github.com/PyCQA/pylint-django/blob/master/pylint_django/augmentations/__init__.py#L846) and the checker and hence it was throwing exception.

### DIFF
--- a/pylint_django/augmentations/__init__.py
+++ b/pylint_django/augmentations/__init__.py
@@ -1,5 +1,6 @@
 """Augmentations."""
 #  pylint: disable=invalid-name
+import functools
 import itertools
 
 from astroid import InferenceError
@@ -737,6 +738,7 @@ def is_class(class_name):
 
 
 def wrap(orig_method, with_method):
+    @functools.wraps(orig_method)
     def wrap_func(*args, **kwargs):
         with_method(orig_method, *args, **kwargs)
     return wrap_func


atodorov:
@mohi7solanki - can you post a reproducer ?

mohi7solanki:
@atodorov 

<augmentations.py>

`from pylint_plugin_utils import augment_visit
from pylint.checkers.variables import VariablesChecker


def dummy(chain, _):
    chain()


def apply_augmentation(linter):
    augment_visit(linter, VariablesChecker.leave_module, dummy)


def register(linter):
    apply_augmentation(linter)
`

run this plugin with pylint-django
`PYTHONPATH=. pylint --load-plugins=pylint_django,augmentations t.py`

mohi7solanki:
@atodorov ping.

atodorov:
@mohi7solanki I don't understand your reproducer.

We need a code snippet which triggers the error in the current version of pylint-django. Ideally you should add your snippet under `tests/input/` so it will be discovered automatically.

As of now we're not seeing exceptions in the existing test suite so I'm not willing to merge this patch b/c I don't really understand the reason for it nor the exact change in behavior.

mohi7solanki:
@atodorov I have created a pylint plugin where I am using pylint-plugin-utils, If I run pylint with both pylint-django and my pylint plugin then it breaks, because I am augmenting the `VariablesChecker.leave_module` and in pylint-django [here](https://github.com/PyCQA/pylint-django/blob/master/pylint_django/augmentations/__init__.py#L848) you are wrapping the leave_module in [wrap](https://github.com/PyCQA/pylint-django/blob/master/pylint_django/augmentations/__init__.py#L741) which doesn't preserve the original method's info like `__name__`, `__qualname__`, `__module__` etc but this is vital for `augment_visit` in pylint-plugin-utils  to work [here](https://github.com/PyCQA/pylint-plugin-utils/blob/43fde2104fb4e7d6aa3a6a71725bca31d44eb936/pylint_plugin_utils/__init__.py#L50).

PR #234, opened 2019-05-23
Fix unbound local error for fk

imomaliev:
Fixes #232

Had to create test in `models` package to check for https://github.com/PyCQA/pylint-django/issues/232#issuecomment-491367551 usecase

Issue #232, opened 2019-04-26
UnboundLocalError: local variable 'key_cls' referenced before assignment


federicobond:
I found this failure in my CI logs, but could not reproduce it locally. The corresponding code looks wrong but I don't know exactly how to correct it. If I can be of any help fixing it let me know.

````
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/astroid/__init__.py", line 91, in _inference_tip_cached
    return iter(_cache[func, node])
KeyError: (<function infer_key_classes at 0x7fe149cfe048>, <Call l.306 at 0x7fe142e66320>)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/bin/pylint", line 11, in <module>
    sys.exit(run_pylint())
  File "/usr/local/lib/python3.6/site-packages/pylint/__init__.py", line 20, in run_pylint
    Run(sys.argv[1:])
  File "/usr/local/lib/python3.6/site-packages/pylint/lint.py", line 1628, in __init__
    linter.check(args)
  File "/usr/local/lib/python3.6/site-packages/pylint/lint.py", line 943, in check
    self._do_check(files_or_modules)
  File "/usr/local/lib/python3.6/site-packages/pylint/lint.py", line 1075, in _do_check
    self.check_astroid_module(ast_node, walker, rawcheckers, tokencheckers)
  File "/usr/local/lib/python3.6/site-packages/pylint/lint.py", line 1158, in check_astroid_module
    walker.walk(ast_node)
  File "/usr/local/lib/python3.6/site-packages/pylint/utils.py", line 1303, in walk
    self.walk(child)
  File "/usr/local/lib/python3.6/site-packages/pylint/utils.py", line 1300, in walk
    cb(astroid)
  File "/usr/local/lib/python3.6/site-packages/pylint_plugin_utils/__init__.py", line 57, in augment_func
    augmentation(chain, node)
  File "/usr/local/lib/python3.6/site-packages/pylint_plugin_utils/__init__.py", line 148, in do_suppress
    chain()
  File "/usr/local/lib/python3.6/site-packages/pylint_plugin_utils/__init__.py", line 56, in chain
    old_method(node)
  File "/usr/local/lib/python3.6/site-packages/pylint_plugin_utils/__init__.py", line 57, in augment_func
    augmentation(chain, node)
[...]

PR #239, opened 2019-07-07
Cherry-pick PR #234 with more documentation and passing tests

atodorov:
@imomaliev I've cherry-picked your changes from #234 and added a pylint disable=no-member for the test to pass. The reason for that is b/c your patch only solves the traceback reported in #232 so the test will check only for that.

See also https://github.com/PyCQA/pylint-django/issues/232#issuecomment-495242695. The existing code base tries to be smart about which modules it loads before searching for model classes and not everything is loaded. I'm not even sure we can handle all cases or if we want to make the code even more complicated than it is.

The whole reason somebody needs to reference a FK model by string is circular dependencies in their own code which means they have to improve it in the first place. Which is something pylint is designed to help you with, not jumping through hoops so it is able to discover all of your files.

Once tests here pass I will merge and release a new version.

### DIFF
--- a/pylint_django/tests/input/models/func_noerror_foreign_key_key_cls_unbound_in_same_package.py
+++ b/pylint_django/tests/input/models/func_noerror_foreign_key_key_cls_unbound_in_same_package.py
@@ -1,0 +1,15 @@
+"""
+Checks that Pylint does not complain about ForeignKey pointing to model
+in module of models package
+"""
+# pylint: disable=missing-docstring
+from django.db import models
+
+
+class FairyTail(models.Model):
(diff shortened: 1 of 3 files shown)


imomaliev:
@atodorov I am not sure that silencing "no-member" is way to go. Because, yes, there won't be any exception after my changes, but pylint still wouldn't understand that `self.my_fk.field` is correct lookup for case like this `ForeignKey('Author')`. Maybe we should add warning that all string values for `ForeignKey` should comply to pattern `app.Model` this way it will be easier for us to introspect dependencies and also it is better and more explicit model definition style

atodorov:
Warning sounds good but eitherway it is a separate functionality.

PR #252, opened 2019-10-30
Fix tests by using pylint/tests/test_functional from a git clone

atodorov:


atodorov:
@PCManticore `LintModuleTest` and `FunctionalTestFile` is what pylint-django test suite uses. If you can move these to testutils, great. I can even do so myself and send a PR in the next few days.

FTR I will probably merge this PR as-is b/c we've got another PR open which needs to run through the tests before I can release a new version.

carlio:
FWIW the tests haven't changed in architecture for several years so I'm all for refreshing them :-)

PR #256, opened 2019-11-24
Do not rely on the version number to know if the tests are present and fixing test matrix

aerostitch:
Hi guys,

Sometimes the distro re-add the files that were removed from the pip packages and in this case it will probably be the case in a few days, so maybe it would be better to check if the test directory exists instead of relying on the version. What do you think?

Thanks for your help,
Joseph

PR #257, opened 2019-11-30
Another take on PR #256 so we can have tests working again

atodorov:
@aerostitch I've taken your changes from #256 and squashed related commits together and removed some things and updated the order of operations. 

- try to import test classes from `pylint.testutils` - will work once we have 2.5 released
- for Travis CI just git clone the 2.4 branch from which we can import `tests_functional`. The code will try to use `PYLINT_TEST_FUNCTIONAL_DIR` but will default to the Travis location
- if that fails try to use an older location

The only reason you may need these changes is if you are trying to either develop/test pylint-django locally or maybe you are trying to build a binary package for a Linux distro. The changes above should make it possible to workaround the current state of affairs.

Or alternatively install pylint from master branch and use that instead.

Tell me what you think ?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ matrix:
     - { stage: build_and_package_sanity, python: 3.6, env: SANITY_CHECK=1 }
 
 before_install:
-  - git clone --depth 1 https://github.com/PyCQA/pylint.git ~/pylint
+  - git clone --depth 1 https://github.com/PyCQA/pylint.git --branch 2.4 --single-branch ~/pylint
 
 install:
   - pip install tox-travis
--- a/pylint_django/tests/input/func_noerror_foreign_key_package.py
+++ b/pylint_django/tests/input/func_noerror_foreign_key_package.py
@@ -7,7 +7,7 @@ from django.db import models
 
 
 class Book(models.Model):
-    author = models.ForeignKey(to='input.Author', on_delete=models.CASCADE)
+    author = models.ForeignKey(to='pylint_django.tests.input.Author', on_delete=models.CASCADE)
 
     def get_author_name(self):
         return self.author.id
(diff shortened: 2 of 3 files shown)


aerostitch:
Hi @atodorov!
The PR doesn't fix the problem of incorrect tox configuration so the test against the django-master tox target is kinda useless as-is and any specific tests where we try to overwrite package versions, especially for pylint will be prone to issues due to the multi-versioning of dependencies installed halfway on the system level and halfway in the tox virtual environment. It would be good to address these issues sooner than later to avoid problems in the future (maybe in a separate PR?) IMHO.
But otherwise it looks good.
Thanks.
Joseph

aerostitch:
The reason I say django-master is useless is that it tests against pylint 2.4 (system-level package) instead of pylint master as configured in the tox env. You can see more details there: https://github.com/PyCQA/pylint-django/pull/256#issuecomment-557968578

atodorov:
@aerostitch just so that we're on the same page. Was your original intention behind #256 to fix the ste suite running against django-master or something else?

aerostitch:
No it wasn't but you asked me to fix the test suite in that PR and that's where I found the tox issue which is part of fixing the test suite. Right now your tests pasd on that tox config only because you're testing against the wrong package version. #256 would have had a lot less commits if I'd had stick to the original intent.

atodorov:
> No it wasn't but you asked me to fix the test suite in that PR and that's where I found the tox issue which is part of fixing the test suite.

@aerostitch right, I did ask you to fix an unrelated issue but the original PR became too big. As for testing against django-master+pylint-master I will need to experiment a bit first. However this is outside of the current scope.

Above you said this PR looks good. Does it fix your original problem that distros add back the test files to the package ?  If yes, let me know if you prefer to merge #256 (minus the django-master changes) or if I should go ahead and merge this one?

aerostitch:
@atodorov the merge of either works for me. Thanks!

Issue #80, opened 2016-07-20
ugettext_lazy('').format


sassanh:
Thanks for this great extension.
Currently `pylint` with `pylint-django` returns error `pylint: [no-member] Instance of '__proxy__' has no 'format' member` for something like `ugettet_lazy('').format`.
I fixed it by changing `ugettext_lazy = lambda x: None` to `ugettext_lazy = lambda x: ''` in file `pylint_django/transforms/transforms/django_utils_translation.py`. Is it alright?

PR #262, opened 2020-03-31
Fix error for gettext_lazy

Anto59290:
Hey :)

ugettext_lazy will soon be removed in favour of gettext_lazy. See https://docs.djangoproject.com/en/3.0/releases/3.0/#id3 for more details.

This fix a very similar issue to https://github.com/PyCQA/pylint-django/issues/80. We (with @rik) also added a test.

### DIFF
--- a/pylint_django/transforms/transforms/django_utils_translation.py
+++ b/pylint_django/transforms/transforms/django_utils_translation.py
@@ -1,2 +1,5 @@
-def ugettext_lazy(_):
+def gettext_lazy(_):
     return ''
+
+
+ugettext_lazy = gettext_lazy  # pylint:disable=invalid-name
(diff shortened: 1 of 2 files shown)


pylint_django/transforms/transforms/django_utils_translation.py
```diff
@@ -1,2 +1,5 @@
-def ugettext_lazy(_):
+def gettext_lazy(_):
     return ''
+
+
+ugettext_lazy = gettext_lazy
```
atodorov, line 5: can you disable the invalid-name pylint check here. pylint expects this to be all caps because it doesn't understand this is supposed to be a function.

Anto59290:
Thanks :pray:

PR #270, opened 2020-07-06
New linter: warn about missing backwards migrations

brymut:
Refs kiwitcms/Kiwi#1774 New linter: warn about missing backwards migrations.

@atodorov, not quite sure how to handle/structure the testing for this. Should I add migrations to test with on top of the tests for the db_performance checker or should I have it with the other checkers? As it looks like the db_performance is added on separately to add the migrations path?

### DIFF
--- a/pylint_django/tests/input/migrations/0003_without_backwards.txt
+++ b/pylint_django/tests/input/migrations/0003_without_backwards.txt
@@ -1,0 +1,4 @@
+missing-backwards-migration-callable:13:Migration:pylint_django.tests.input.migrations.0003_without_backwards Always include backwards migration callable
+missing-backwards-migration-callable:14:Migration:pylint_django.tests.input.migrations.0003_without_backwards Always include backwards migration callable
+missing-backwards-migration-callable:16:Migration:pylint_django.tests.input.migrations.0003_without_backwards Always include backwards migration callable
+missing-backwards-migration-callable:18:Migration:pylint_django.tests.input.migrations.0003_without_backwards Always include backwards migration callable
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -15,4 +15,5 @@
 * [psrb](https://github.com/psrb)
 * [WayneLambert](https://github.com/WayneLambert)
 * [alejandro-angulo](https://github.com/alejandro-angulo)
+* [brymut](https://github.com/brymut)
 
--- a/pylint_django/checkers/db_performance.py
+++ b/pylint_django/checkers/db_performance.py
@@ -14,6 +14,7 @@ from pylint import checkers
 from pylint.checkers import utils
 from pylint_django.__pkginfo__ import BASE_ID
 from pylint_django import compat
+from pylint_django.utils import is_migrations_module
 
 
 def _is_addfield_with_default(call):
@@ -35,13 +36,6 @@ def _is_addfield_with_default(call):
                     return True
 
     return False
-
-
-def _is_migrations_module(node):
-    if not isinstance(node, astroid.Module):
-        return False
-
-    return 'migrations' in node.path[0] and not node.path[0].endswith('__init__.py')
 
 
 class NewDbFieldWithDefaultChecker(checkers.BaseChecker):
@@ -69,7 +63,7 @@ class NewDbFieldWithDefaultChecker(checkers.BaseChecker):
     _possible_offences = {}
 
     def visit_module(self, node):
-        if _is_migrations_module(node):
+        if is_migrations_module(node):
             self._migration_modules.append(node)
 
     def visit_call(self, node):
@@ -78,7 +72,7 @@ class NewDbFieldWithDefaultChecker(checkers.BaseChecker):
         except:  # noqa: E722, pylint: disable=bare-except
             return
 
-        if not _is_migrations_module(module):
+        if not is_migrations_module(module):
             return
 
         if _is_addfield_with_default(node):
@@ -114,6 +108,38 @@ class NewDbFieldWithDefaultChecker(checkers.BaseChecker):
                     self.add_message('new-db-field-with-default', args=module.name, node=node)
 
 
+class MissingBackwardsMigrationChecker(checkers.BaseChecker):
+    __implements__ = (interfaces.IAstroidChecker,)
+
+    name = 'missing-backwards-migration-callable'
+
+    msgs = {'W%s05' % BASE_ID: ('%s Always include backwards migration callable',
+                                'missing-backwards-migration-callable',
+                                'Always include a backwards/reverse callable counterpart'
+                                ' so that the migration is not irreversable.')}
+
+    @utils.check_messages('missing-backwards-migration-callable')
+    def visit_call(self, node):
+        try:
+            module = node.frame().parent
+        except:  # noqa: E722, pylint: disable=bare-except
+            return
+
+        if not is_migrations_module(module):
+            return
+
+        if node.func.as_string().endswith('RunPython') and len(node.args) < 2:
+            if node.keywords:
+                for keyword in node.keywords:
+                    if keyword.arg == 'reverse_code':
+                        return
+                self.add_message('missing-backwards-migration-callable',
+                                 args=module.name, node=node)
+            else:
+                self.add_message('missing-backwards-migration-callable',
+                                 args=module.name, node=node)
+
+
 def load_configuration(linter):
     # don't blacklist migrations for this checker
     new_black_list = list(linter.config.black_list)
@@ -125,5 +151,6 @@ def load_configuration(linter):
 def register(linter):
     """Required method to auto register this checker."""
     linter.register_checker(NewDbFieldWithDefaultChecker(linter))
+    linter.register_checker(MissingBackwardsMigrationChecker(linter))
     if not compat.LOAD_CONFIGURATION_SUPPORTED:
         load_configuration(linter)
--- a/pylint_django/tests/input/migrations/0003_without_backwards.py
+++ b/pylint_django/tests/input/migrations/0003_without_backwards.py
@@ -1,0 +1,20 @@
+# pylint: disable=missing-docstring, invalid-name
+from django.db import migrations
+
+
+# pylint: disable=unused-argument
+def forwards_test(apps, schema_editor):
+    pass
+
+
+class Migration(migrations.Migration):
+
+    operations = [
+        migrations.RunPython(),  # [missing-backwards-migration-callable]
+        migrations.RunPython(  # [missing-backwards-migration-callable]
+            forwards_test),
+        migrations.RunPython(  # [missing-backwards-migration-callable]
+            code=forwards_test),
+        migrations.RunPython(  # [missing-backwards-migration-callable]
+            code=forwards_test, atomic=False)
+    ]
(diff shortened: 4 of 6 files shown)


pylint_django/checkers/__init__.py
```diff
@@ -11,3 +12,4 @@ def register_checkers(linter):
     linter.register_checker(DjangoInstalledChecker(linter))
     linter.register_checker(JsonResponseChecker(linter))
     linter.register_checker(FormChecker(linter))
+    linter.register_checker(MissingBackwardsMigrationChecker(linter))
```
atodorov, line 15: @carlio this whole PR is for something I've identified in a project of mine. However I'm wondering whether or not we want this entire checker to be enabled by default or have people enabled it manually as an extension. What do you think? 

NB: depending on the answer this PR may need some adjustments like moving duplicate functions around, renaming modules, etc.

pylint_django/checkers/migrations.py
```diff
@@ -0,0 +1,41 @@
+import astroid
+from pylint import interfaces
+from pylint.checkers import utils, BaseChecker
+from pylint_django.__pkginfo__ import BASE_ID
+
+def _is_migrations_module(node):
+    if not isinstance(node, astroid.Module):
+        return False
+
+    return 'migrations' in node.path[0] and not node.path[0].endswith('__init__.py')
```
atodorov, line 10: Note: this function is duplicate between here and db_performance.py so it would probably need to be moved to a common place. I think utils.py is a good place.

pylint_django/checkers/migrations.py
```diff
@@ -0,0 +1,41 @@
...
+    def visit_call(self, node):
+        try:
+            module = node.frame().parent
+        except:  # noqa: E722, pylint: disable=bare-except
+            return
+
+        if not _is_migrations_module(module):
+            return
+
+        if node.func.as_string().endswith('RunPython') and len(node.args) < 2:
```
atodorov, line 32: Don't use `as_string()`, function calls have attribute which has a name which matches `RunPython`.

pylint_django/checkers/migrations.py
```diff
@@ -0,0 +1,41 @@
...
+            if node.keywords:
+                for keyword in node.keywords:
+                    if keyword.arg == 'reverse_code':
+                        return
```
atodorov, line 36: I'm not sure what this block does, can you explain ?

pylint_django/checkers/__init__.py
atodorov: Note2: IMO this should go together with `db_performance.py` (and we rename the module) because the test suite is structured around 2 items: tests for regular checkers and tests for checkers around migrations which are explicitly enabled. I woudn't want to duplicate that if possible.

pylint_django/tests/input/migrations/0004_with_backwards.py
```diff
@@ -0,0 +1,17 @@
...
+
+def backwards_test(apps, schema_editor):
+    pass
+
+
+class Migration(migrations.Migration):
+
+    operations = [
+        migrations.RunPython(forwards_test, backwards_test)
+    ]
```
atodorov, line 17: see `test_db_performance_plugin()` inside `pylint_django/tests/test_func.py` for hints. If you want to experiment as-is you will have to duplicate the test function b/c it needs the additional file paths. Migrations are automatically excluded by the main pylint-django plugin, see plugin.py -> load_configuration() -> linter.config.black_list

atodorov:
@brymut also rebase to the latest master which fixed the issue with failing tests in CI

pylint_django/checkers/migrations.py
brymut: the reverse handler can also be passed into [`RunPython`](https://docs.djangoproject.com/en/3.0/ref/migration-operations/#runpython) via keyword argument `reverse_code` like in https://github.com/kiwitcms/Kiwi/blob/d7e456e4d56b15f262a7bade89bc90036590fc93/tcms/core/migrations/0001_squashed.py#L62 if its not via pos arg. So this is just for the checker to confirm if it is in the keyword arguments

pylint_django/checkers/migrations.py
brymut: I had tried `node.func.attrname` but I was getting AttributeError's. I think we might run into the same issue as [here](https://github.com/kiwitcms/Kiwi/pull/1742#discussion_r441485689) if we rely on the attribute of the node to give us the name. We would have to assume that `RunPython` isn't being imported directly as opposed to using the whole migrations import.

pylint_django/tests/input/migrations/0004_with_backwards.py
brymut: Will wait to see which structure for this checker will be used.

pylint_django/checkers/__init__.py
atodorov: @brymut I have made up my mind. Move the new checker class inside the `db_performance` module. That will make it easier with the automation tests as well. After we merge this PR I will take care of renaming things and updating the docs.

pylint_django/checkers/migrations.py
atodorov: This check needs to be more strict. Use isinstance() to make sure that node.func is an Attribute and then you can safely access node.func.attrname

pylint_django/checkers/migrations.py
atodorov: I see. This should be more robust then. Because I can also specify `code=forwards` as a keyword and then node.args becomes 0 length.

Here's the full signature of RunPython:
https://docs.djangoproject.com/en/3.0/ref/migration-operations/#django.db.migrations.operations.RunPython

brymut: Wouldn't this check already catch that? I already had this in mind with the example I linked from the Kiwi repo.

pylint_django/checkers/migrations.py
brymut: But what if 
`from django.db.migrations import RunPython` is used?
wouldn't `RunPython(forwards, backwards)` be a `Name` node type? and be missed by the winter?

carlio: One slight problem we've run into before is that any imports from `django.db` make django try to configure the DB connection and this can fail, especially when Django isn't installed for CI jobs (which happens annoyingly more than you'd expect).

pylint_django/checkers/__init__.py
carlio: @atodorov I agree I think it makes sense to group it together with the `db_performance` module as it seems related but I'm easy either way.

brymut:
@atodorov may I please get some help, I can't seem to be able to get the test to recognise the `0003_without_backwards.txt` to compare the expected error output against like the `0002_new_column.txt`

pylint_django/checkers/migrations.py
atodorov: > One slight problem we've run into before is that any imports from django.db make django try to configure the DB connection and this can fail, especially when Django isn't installed for CI jobs (which happens annoyingly more than you'd expect).

@carlio you have misread the comment from @brymut. He meant that when the migration imports RunPython as `from django.db.migrations import RunPython` it changes the astroid node type to Name (because it is a method/class name) and when it's used as `migrations.RunPython` it is an attribute b/c we're accessing an attribute of the migrations module.

atodorov: @brymut agreed, let's leave the as_string() comparison although I don't like it. 

FTR in the reference that you posted the string comparison has an unwanted side effect because forms, models and admin classes can have fields whose name ends with `Field`. And their types are different hence the broader match. 

For RunPython this is less so.

pylint_django/checkers/migrations.py
atodorov: Nope. I can write my migration as such:
```
migrations.RunPython(code=forwards, reverse_code=backwards)
```

and node.args will be an empty list. However this is perfectly valid invocation.

Or I can call it like so:

```
migrations.RunPython(code=forwards, atomic=False)
```

and node.args will still be empty list but this invocation is missing the reverse_code argument.

atodorov: IMO a break-down of the if condition will simplify things, e.g.

1) if len(node.args) < 2 - definitely an error condition, report it and return from the method
2) if not then inspect kwargs for both code/reverse_code keywords taking into account that the first one can also be a positional argument.

pylint_django/tests/input/migrations/0003_without_backwards.txt
```diff
@@ -0,0 +1 @@
+missing-backwards-migration-callable:12:Migration:pylint_django.tests.input.migrations.0003_without_backwards Always include backwards migration callable
```
atodorov, line 1: The error in CI that you are seeing with this test is that the txt file tells it to expect the error on line 12 while the RunPyhon node which gets inspected and then used for error annotations in on line 13. See the log message:
```
Unexpected in testdata:

E             13: missing-backwards-migration-callable
```

pylint_django/tests/input/migrations/0004_with_backwards.py
```diff
@@ -0,0 +1,19 @@
+# pylint: disable=missing-docstring, invalid-name
```
atodorov, line 1: filename should be `0004_noerror_with_backwards.py`

pylint_django/checkers/migrations.py
brymut: I've double-checked with your example, the linter will still catch it. The for loop on line 133 still won't find the reverse_code argument and the linter will bring up the issue.

pylint_django/tests/input/migrations/0004_with_backwards.py
```diff
@@ -0,0 +1,19 @@
...
+# pylint: disable=unused-argument
+def backwards_test(apps, schema_editor):
+    pass
+
+
+class Migration(migrations.Migration):
+
+    operations = [
+        migrations.RunPython(forwards_test, backwards_test)
+    ]
```
atodorov, line 19: @brymut you also need to provide  all of the calling variantions, incl. only as `RunPython()` to make sure all of the conditionals in the checker work as expected.

pylint_django/checkers/migrations.py
atodorov: Add a test case for it then.

pylint_django/tests/input/migrations/0003_without_backwards.txt
brymut: Can't believe I missed that. I wasn't paying attention to that, don't know why I was fixated on 12. Thank you

brymut: Still having the same problem for some reason.

pylint_django/checkers/migrations.py
brymut: Added your examples and the linter is working as expected.

pylint_django/tests/input/migrations/0003_without_backwards.txt
atodorov: Every line in the test input file which is doing to trigger an error also needs to be annotated with a comment to tell the test suite what error message is expected, see `0002_new_column.py`. Then this is matched against the actual messages received from the linter.

Issue #243, opened 2019-08-06
RFC: new checker: ForeignKey models referenced by string


atodorov:
Given what we have in our Known issues section (in README) I propose that by default pylint-django starts reporting on FK and any other related field types which reference the related model via string constant not its class. 

pylint-django probes for some relatively simple and rather common locations when trying to search for related models, documented in #242. However we can't and will not try to probe other possible locations because this will turn into maintenance hell (patching paths and expectations) and is not really the scope of this tool IMO. This will (hopefully) minimize the number of bug reports and the general misunderstanding about these types of issues.

I am opening this issue to serve as a place for comments, :+1: and :-1: reactions to see what other people think.

imomaliev:
I use string constants by default. They are much better and cleaner way to specify models without circular import issues. I suggest reporting strings that do not match this pattern `'app.Model'`

carlio:
I think this makes sense, as long as the checked uses a new error code then people can filter that out by ignoring it in their pylint config. I certainly think that it's worth suggesting that imports are used over strings where possible so this is a :+1: from me.

(I also personally think that using strings to avoid circular imports can point to a larger problem of project architecture but that is outside the scope of pylint-django!)

atodorov:
@imomaliev I am with @carlio on this one. However a comment here:

> I suggest reporting strings that do not match this pattern 'app.Model'

The problem is not the string pattern but how the application is structured. For example in #241 the FK reference follows the 'app.Model' pattern, it is `contenttypes.ContentType`. But the actual import to be able to find this model is not `contenttypes.models.ContentType`, instead it is `django.contrib.contenttypes.models.ContentType`.

In my own application we've got `linkreference.LinkReference` where the actual module is `tcms.core.contrib.linkreference.models`.

imomaliev:
@atodorov The actual import can be resoled by reusing django's internal resolution methods. I think there should be a default way for most users. I am not against on removing this complexity from this plugin. Maybe for my case there should be a separate plugin that will do resolution proprely

imomaliev:
@atodorov At least for backward compat we should move current string resolution move to separate module, so the people who are using this functionality in current state won't lose it

rik:
[...]

PR #272, opened 2020-07-12
Handle ForeignKey String References

alejandro-angulo:
Possible solution to https://github.com/PyCQA/pylint-django/issues/243 . This change leverages Django's model name resolution to handle ForeignKey string references.


README.rst
```diff
@@ -95,9 +95,10 @@ This plugin is disabled by default! To enable it::
 Known issues
 ------------
 
-If you reference foreign-key models by their name (as string) ``pylint-django`` may not be
-able to find the model and will report issues because it has no idea what the underlying
-type of this field is. Supported options are::
+If Django is not installed and  you reference foreign-key models by their name (as
+string) ``pylint-django`` may not be able to find the model and will report issues
+because it has no idea what the underlying type of this field is. Supported options
+are::
```
atodorov, line 101: Why the change here ? I don't see anything different in the text.

pylint_django/transforms/foreignkey.py
```diff
@@ -87,10 +100,24 @@ def infer_key_classes(node, context=None):
...
+                use_django_model_resolution = bool(django_spec)
+
+                if use_django_model_resolution:
+                    # If Django is installed we can use it to resolve the module name
+                    try:
+                        module_name = _module_name_from_django_model_resolution(model_name, module_name)
+                    except LookupError:
+                        use_django_model_resolution = False
+
+                if not use_django_model_resolution:
```
atodorov, line 115: that's an `else` to the above block, isn't it ?

pylint_django/tests/test_app/apps.py
```diff
@@ -0,0 +1,5 @@
+from django.apps import AppConfig
+
+
+class TestAppConfig(AppConfig):
+    name = 'test_app'
```
atodorov, line 5: I believe all of this should be under `input/`. I can see a `models/author.py` module in there.

pylint_django/tests/test_func.py
```diff
@@ -6,6 +6,8 @@
 import pylint
 
 
+os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'pylint_django.tests.conftest')
+
```
atodorov, line 10: Why do we need this? checkers/django_installed.py also tries to import Django and it used to work without anything else so please explain.

pylint_django/transforms/foreignkey.py
```diff
@@ -87,10 +100,24 @@ def infer_key_classes(node, context=None):
...
             elif not module_name.endswith('models'):
                 # otherwise Django allows specifying an app name first, e.g.
-                # ForeignKey('auth.User') so we try to convert that to
-                # 'auth.models', 'User' which works nicely with the `endswith()`
-                # comparison below
-                module_name += '.models'
+                # ForeignKey('auth.User')
+
+                django_spec = find_spec('django')
+                use_django_model_resolution = bool(django_spec)
```
atodorov, line 106: is this making sure that Django is installed ? If yes, then you don't need it b/c there's checkers/django_installed.py which will produce a fatal failure if you try executing pylint-django in environment which doesn't have Django.

pylint_django/transforms/foreignkey.py
alejandro-angulo: Yes, but I didn't use an `else` because I wanted to run this logic if django's resolution failed. The value of `use_django_model_resolution` is updated on line 113, inside of the `if` block above.

pylint_django/tests/test_app/apps.py
alejandro-angulo: Ah okay, I wasn't sure if this belonged in there since it's moreso setup than something to test. I'll move this into `input/`.

pylint_django/transforms/foreignkey.py
alejandro-angulo: Yeah this is just a check to see if Django is installed. I didn't know about that check. I'll remove this.

README.rst
alejandro-angulo: I added some additional text to the first line `If Django is not installed...`. But this was before I knew about `checkers/django_installed.py` so I'll undo this change (don't think it makes much sense anymore).

pylint_django/tests/test_func.py
alejandro-angulo: I added this so that the test configuration would be used when `django.setup` is called in `_module_name_from_django_model_resolution` (https://github.com/PyCQA/pylint-django/pull/272/files#diff-1bbed731b242312a86f3f314c3682343R47). But looking back on it, I think it'd be better to call `settings.configure` before `django.setup` to avoid messing with environment variables.

README.rst
alejandro-angulo: I undid my earlier change but maybe the entire known issues section should be remove if this is accepted.

pylint_django/transforms/foreignkey.py
alejandro-angulo: No longer checking to see if Django is installed so these specific ifs aren't needed anymore. Resolving conversation.

pylint_django/tests/test_func.py
atodorov: Makes sense then. Let's name this module settings.py though.

pylint_django/tests/conftest.py
```diff
@@ -0,0 +1,12 @@
+import os
+
+# Build paths inside the project like this: os.path.join(BASE_DIR, ...)
+BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
```
atodorov, line 4: This variable and the import above don't seem to be needed.

pylint_django/transforms/foreignkey.py
```diff
@@ -41,6 +41,18 @@ def _get_model_class_defs_from_module(module, model_name, module_name):
     return class_defs
 
 
+def _module_name_from_django_model_resolution(model_name, module_name):
+    import django  # pylint: disable=import-outside-toplevel
```
atodorov, line 45: IMO this can be a top-level import, no ?

pylint_django/transforms/foreignkey.py
```diff
@@ -41,6 +41,18 @@ def _get_model_class_defs_from_module(module, model_name, module_name):
...
+    django.setup()
+    from django.apps import apps  # pylint: disable=import-outside-toplevel
```
atodorov, line 47: same for this one, top-level ?

pylint_django/transforms/foreignkey.py
```diff
@@ -41,6 +41,18 @@ def _get_model_class_defs_from_module(module, model_name, module_name):
...
+
+    app = apps.get_app_config(module_name)
+    model = app.get_model(model_name)
+    module_name = model.__module__
```
atodorov, line 51: can directly return this value instead of using an intermediate variable.

pylint_django/tests/input/func_noerror_string_foreignkey.py
```diff
@@ -0,0 +1,10 @@
+"""
+Checks that PyLint correctly handles string foreign keys
+https://github.com/PyCQA/pylint-django/issues/243
+"""
+#  pylint: disable=missing-docstring
+from django.db import models
+
+
+class Book(models.Model):
+    author = models.ForeignKey("test_app.Author", models.CASCADE)
```
atodorov, line 10: Let's also add another FK reference to "auth.User" - because django apps don't follow the usual convention of app name vs. module name. That's was one of the reasons we didn't deal with string FKs for so long.

pylint_django/tests/conftest.py
alejandro-angulo: Oh good catch. This was left over from before when I was checking to see if Django was installed. Will fix.

pylint_django/transforms/foreignkey.py
alejandro-angulo: Oh good catch. This was left over from before when I was checking to see if Django was installed. Will fix.

pylint_django/tests/conftest.py
alejandro-angulo: Good point. Don't think `DEBUG` is needed either.

pylint_django/tests/settings.py
```diff
@@ -0,0 +1,12 @@
+SECRET_KEY = 'fake-key'
+
+INSTALLED_APPS = [
+    'django.contrib.auth',
```
alejandro-angulo, line 4: These two apps and the two middleware are required for Django's authentication features (e.g. `auth.User`): https://docs.djangoproject.com/en/3.0/topics/auth/#installation .

pylint_django/transforms/foreignkey.py
```diff
@@ -41,6 +41,17 @@ def _get_model_class_defs_from_module(module, model_name, module_name):
     return class_defs
 
 
+def _module_name_from_django_model_resolution(model_name, module_name):
+    import django  # pylint: disable=import-outside-toplevel
+    django.setup()
+    from django.apps import apps  # pylint: disable=import-outside-toplevel
```
alejandro-angulo, line 47: Moving the imports into the top-level broke the build during the `django_not_installed` check: https://travis-ci.org/github/PyCQA/pylint-django/builds/710245712 .

ajhodges:
https://github.com/PyCQA/pylint-django#known-issues

I think the Readme can probably be updated now 👍👍

515hikaru:
@alejandro-angulo  I was just in trouble facing this problem. Thank you for fixing it!

PR #283, opened 2020-08-04
Changelog for v2.3.0

atodorov:
@carlio I am reluctant to call this 3.0 b/c I expect pylint will move to v3 at some point, likely introducing incompatibilities and we'll probably have a version mismatch at that point. 

It's fine by me to not precisely follow SemVer every time. I can release a new build today if you agree with that.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,16 @@
 Changelog
 =========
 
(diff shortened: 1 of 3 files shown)


carlio:
@atodorov Yeah I agree it makes sense to keep in sync with pylint if there's going to be a new major release soon.

PR #285, opened 2020-08-06
Update `forms.py` to fix #284

### LINKED ISSUE #284: AttributeError: 'Subscript' object has no attribute 'name'
Dear all,

It's been a while I've experienced an issue when running `pylint` with `pylint_django`. I include the error traceback and the line I think causes the problem.

Traceback:
```
  File "D:\DevTools\Python38-32\lib\runpy.py", line 193, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "D:\DevTools\Python38-32\lib\runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "D:\DevRepo\dex\venv\Scripts\pylint.exe\__main__.py", line 9, in <module>
  File "d:\devrepo\dex\venv\lib\site-packages\pylint\__init__.py", line 22, in run_pylint
    PylintRun(sys.argv[1:])
  File "d:\devrepo\dex\venv\lib\site-packages\pylint\lint\run.py", line 344, in __init__
    linter.check(args)
  File "d:\devrepo\dex\venv\lib\site-packages\pylint\lint\pylinter.py", line 870, in check
    self._check_files(
  File "d:\devrepo\dex\venv\lib\site-packages\pylint\lint\pylinter.py", line 904, in _check_files
    self._check_file(get_ast, check_astroid_module, name, filepath, modname)
  File "d:\devrepo\dex\venv\lib\site-packages\pylint\lint\pylinter.py", line 930, in _check_file
    check_astroid_module(ast_node)
  File "d:\devrepo\dex\venv\lib\site-packages\pylint\lint\pylinter.py", line 1062, in check_astroid_module 
    retval = self._check_astroid_module(
  File "d:\devrepo\dex\venv\lib\site-packages\pylint\lint\pylinter.py", line 1107, in _check_astroid_module
    walker.walk(ast_node)
  File "d:\devrepo\dex\venv\lib\site-packages\pylint\utils\ast_walker.py", line 75, in walk
    self.walk(child)
  File "d:\devrepo\dex\venv\lib\site-packages\pylint\utils\ast_walker.py", line 72, in walk
    callback(astroid)
  File "d:\devrepo\dex\venv\lib\site-packages\pylint_django\checkers\forms.py", line 47, in visit_classdef
    if child.targets[0].name == 'exclude':
AttributeError: 'Subscript' object has no attribute 'name'
```

After completely cleaned the file `forms.py` giving the error I think I've managed to isolate the problematic line. The file, still giving the error, now contains only this:

```
from django import forms

from dex import models


class DataSelectionForm(forms.ModelForm):
    """
    Form
    """
    class Meta:
        model = models.DataSelection
        fields = [
            "field",
        ]
        widgets = {}
        widgets["field"] = forms.CheckboxSelectMultiple

```

Of course, I've overly simplified that form class. It seems the error is caused by `widgets["field"] = forms.CheckboxSelectMultiple`, which in the real code is adding a different widget to a specific field after a loop creates the `widgets` dict.

If I use instead `widgets = {"field": forms.CheckboxSelectMultiple}` then everything is fine.

A few details about my configuration:

 - OS: Windows 10 Pro x64
 - Pip: 19.2.3
 - Python: 3.8.2

Result of `pip freeze`:

```
astroid==2.4.2
certifi==2020.6.20
chardet==3.0.4
colorama==0.4.3
Django==2.2.14
django-activity-stream==0.9.0
django-annoying==0.10.6
django-filter==2.3.0
django-formtools==2.2
django-jsonstore==0.4.1
django-material==1.6.7
django-simple-history==2.11.0
django-viewflow==1.6.1
docutils==0.16
gunicorn==20.0.4
idna==2.10
isort==4.3.21
lazy-object-proxy==1.4.3
mccabe==0.6.1
oauthlib==3.1.0
Pillow==7.2.0
psycopg2==2.8.5
pyasn1==0.4.8
pyasn1-modules==0.2.8
pylint==2.5.3
pylint-django==2.2.0
pylint-plugin-utils==0.6
python-ldap==3.3.1
python-magic==0.4.18
python-magic-bin==0.4.14
pytz==2020.1
PyYAML==5.3.1
requests==2.24.0
requests-oauthlib==1.3.0
rope==0.17.0
six==1.15.0
sqlparse==0.3.1
toml==0.10.1
urllib3==1.25.10
wrapt==1.12.1
```

Thank you for your time!

uy-rrodriguez:
An exception `AttributeError` is raised when `child.targets[0]` does not have an attribute `name`, as stated in the issue #284 .

`meta.get_children()` returns all children of the `Meta` class, which in the sample code contains a dictionary assignment.

The expected children are of type `Assign` whose first item in the attribute `targets` is of type `AssignName`, like `AssignName.widgets(name='widgets')`.

An instruction like `widget["field"] = forms.SomeWidget` will result in a `Assign` child containing a `Subscript(ctx=<Context.Store: 2>, value=<Name.widgets ...>, slice=<Index ...>)`. This is of course unexpected.

I guess other unexpected instructions in the `Meta` class can trigger similar errors, so probably the most economic solution is to enclose the code in `try... except` and ignore the `AttributeError`. Trying to identify the type of object before accessing the `name` attribute would probably incur in a slower execution.

### DIFF
--- a/pylint_django/checkers/forms.py
+++ b/pylint_django/checkers/forms.py
@@ -1,5 +1,5 @@
 """Models."""
-from astroid.nodes import Assign, ClassDef
+from astroid.nodes import Assign, AssignName, ClassDef
 
 from pylint.interfaces import IAstroidChecker
 from pylint.checkers.utils import check_messages
@@ -41,14 +41,10 @@ class FormChecker(BaseChecker):
             return
 
         for child in meta.get_children():
-            if not isinstance(child, Assign):
+            if (not isinstance(child, Assign)
+                    or not isinstance(child.targets[0], AssignName)):
                 continue
 
-            # Capture and ignore AttributeError raised when targets[0] does not
-            # have an attribute "name"
-            try:
-                if child.targets[0].name == 'exclude':
-                    self.add_message('W%s04' % BASE_ID, node=child)
-                    break
-            except AttributeError:
-                pass
+            if child.targets[0].name == 'exclude':
+                self.add_message('W%s04' % BASE_ID, node=child)
+                break
--- a/pylint_django/checkers/forms.py
+++ b/pylint_django/checkers/forms.py
@@ -44,6 +44,11 @@ class FormChecker(BaseChecker):
             if not isinstance(child, Assign):
                 continue
 
-            if child.targets[0].name == 'exclude':
-                self.add_message('W%s04' % BASE_ID, node=child)
-                break
+            # Capture and ignore AttributeError raised when targets[0] does not
+            # have an attribute "name"
+            try:
+                if child.targets[0].name == 'exclude':
+                    self.add_message('W%s04' % BASE_ID, node=child)
+                    break
+            except AttributeError:
+                pass
--- a/pylint_django/tests/input/func_noerror_forms_py33.py
+++ b/pylint_django/tests/input/func_noerror_forms_py33.py
@@ -31,3 +31,12 @@ class TestFormSubclass(forms.Form):
 class TestModelFormSubclass(forms.ModelForm):
     class Meta:
         pass
+
+
+class TestFormWidgetAssignment(forms.Form):
+
+    multi_field = forms.MultipleChoiceField(choices=[('1', 'First'), ('2', 'Second')])
+
+    class Meta:
+        widgets = {}
+        widgets['multi_field'] = forms.CheckboxSelectMultiple


pylint_django/checkers/forms.py
```diff
@@ -44,6 +44,11 @@ def visit_classdef(self, node):
             if not isinstance(child, Assign):
                 continue
 
-            if child.targets[0].name == 'exclude':
-                self.add_message('W%s04' % BASE_ID, node=child)
-                break
+            # Capture and ignore AttributeError raised when targets[0] does not
+            # have an attribute "name"
+            try:
+                if child.targets[0].name == 'exclude':
```
asottile, line 50: a better solution might be to guard this with `isinstance(child.targets[0], WhateverExpectedType) and ...`

uy-rrodriguez: I completely agree, but wouldn't this incur in a longer execution time?

asottile: imo it's better to have that and not hide bugs (`except AttributeError` could hide any number of other things here) -- this could also be added to the `if isinstance(...): continue` at the top possibly

uy-rrodriguez: Actually that line filters the children and the problem is in the `targets` items. But you're right about hiding other problems. I'm on it!

pylint_django/checkers/forms.py
```diff
@@ -41,7 +41,8 @@ def visit_classdef(self, node):
             return
 
         for child in meta.get_children():
-            if not isinstance(child, Assign):
+            if (not isinstance(child, Assign)
+                    or not isinstance(child.targets[0], AssignName)):
```
atodorov, line 45: IMO you should copy the example code from #284 into one of the existing tests. `func_noerror_forms_py33.py` looks like a good candidate.

atodorov (changes requested):

pylint_django/checkers/forms.py
uy-rrodriguez: Hi! I took a look at the test files and got a bit lost, then couldn't find the time to read all the files in detail, so thank you for pointing me out to the best candidate. I'll give it a try.

Issue #286, opened 2020-08-15
Allow passing settings path using argument instead of env


Goldziher:
Hi there,

There is a problem using this plugin together with Pre-Commit because its impossible to inject env variables in the Pre-Commit hooks. I'd like to suggest / request that you guys add an optional argument for the path of settings.py for a given project, or use another scheme to get it. 

Thanks

Issue #277, opened 2020-07-29
Pylint crashes if Django is not configured


ljodal:
Hi there, starting with version 2.2.0 I'm seeing a crash if django is not properly configured. I'm assuming this is originating from https://github.com/PyCQA/pylint-django/issues/243. This happens even if the string foreign key is resolvable without using the Django runtime.

While I kinda understand that you'd expect to have a Django configured while running pylint, it's not always the case and in some cases it might be hard (like in an editor).

```
Traceback (most recent call last):
  File "<snip>/.venv/lib/python3.7/site-packages/astroid/__init__.py", line 93, in _inference_tip_cached
    return iter(_cache[func, node])
KeyError: (<function infer_key_classes at 0x1113d5b00>, <Call l.162 at 0x11b5f3650>)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<snip>/.venv/bin/pylint", line 8, in <module>
    sys.exit(run_pylint())
  File "<snip>/.venv/lib/python3.7/site-packages/pylint/__init__.py", line 22, in run_pylint
    PylintRun(sys.argv[1:])
  File "<snip>/.venv/lib/python3.7/site-packages/pylint/lint/run.py", line 344, in __init__
    linter.check(args)
  File "<snip>/.venv/lib/python3.7/site-packages/pylint/lint/pylinter.py", line 871, in check
    self.get_ast, self._iterate_file_descrs(files_or_modules)
  File "<snip>/.venv/lib/python3.7/site-packages/pylint/lint/pylinter.py", line 904, in _check_files
    self._check_file(get_ast, check_astroid_module, name, filepath, modname)
  File "<snip>/.venv/lib/python3.7/site-packages/pylint/lint/pylinter.py", line 930, in _check_file
    check_astroid_module(ast_node)
  File "<snip>/.venv/lib/python3.7/site-packages/pylint/lint/pylinter.py", line 1063, in check_astroid_module
    ast_node, walker, rawcheckers, tokencheckers
  File "<snip>/.venv/lib/python3.7/site-packages/pylint/lint/pylinter.py", line 1107, in _check_astroid_module
    walker.walk(ast_node)
  File "<snip>/.venv/lib/python3.7/site-packages/pylint/utils/ast_walker.py", line 75, in walk
    self.walk(child)
  File "<snip>/.venv/lib/python3.7/site-packages/pylint/utils/ast_walker.py", line 75, in walk
    self.walk(child)
[...]

PR #287, opened 2020-08-26
Improved Django settings handling

carlio:
This revisits #277 and #243 @atodorov @alejandro-angulo 

I had a new attempt atdealing with django settings required by the foreign key string handling and came up with this solution:

1) as currently, if `DJANGO_SETTINGS_MODULE` is specified, those settings will be used.
2) a commandline option can be used instead - for example `pylint --load-plugins=pylint_django --django-settings-module=myproject.settings` which I assume would work from a `.pylintrc` too.
3) if neither of these are set, then Django is configured using the default settings. In this case, an error message is emitted (`E5110`) instructing the user to configure Django manually to improve linting. I set this as ERROR because it's more severe than just a warning since it is kind of an error in the testing (not in the code being tested, but the testing CI setup). It is not fatal, however, as it can continue to run.

This is done by moving the Foreign Key checks into a checker rather than loading the transforms in when the plugin is loaded. This allows for using the `open()` "setup" method to do any pre-check startup, in this case, running the Django `settings.configure()` method manually rather than allowing Django to do it by itself if it has an environment variable.

Using a checker also means that a commandline option can be used as an additional option for configuring Django settings, which will allow it to be used in pre-commit for example (thus fixing #286)

Note: this also had to remove the `--no-binary :all: ` from the `scripts/build.sh`  sanity check because it seems that a newer release of isort (which pylint depends on) requires clikit and when `--no-binary` is used, pylint-django fails to install. I'm not sure of the consequences of this but I assume that if pylint needs it to install, we have to accept that. This is not related to the main contents of the PR, just something I had to do to get the latest versions to pass CI.

### DIFF
--- a/pylint_django/checkers/django_installed.py
+++ b/pylint_django/checkers/django_installed.py
@@ -26,5 +26,5 @@ class DjangoInstalledChecker(BaseChecker):
         except ImportError:
             # mild hack: this error is added before any modules have been inspected
             # so we need to set some kind of value to prevent the linter failing to it
-            self.linter.set_current_module('project global')
+            self.linter.set_current_module('pylint_django')
             self.add_message('django-not-available')
--- a/pylint_django/checkers/django_installed.py
+++ b/pylint_django/checkers/django_installed.py
@@ -20,8 +20,11 @@ class DjangoInstalledChecker(BaseChecker):
     }
 
     @check_messages('django-not-available')
-    def close(self):
+    def open(self):
         try:
             __import__('django')
         except ImportError:
-            self.add_message('F%s01' % BASE_ID)
+            # mild hack: this error is added before any modules have been inspected
+            # so we need to set some kind of value to prevent the linter failing to it
+            self.linter.set_current_module('project global')
+            self.add_message('django-not-available')
--- a/pylint_django/checkers/foreign_key_strings.py
+++ b/pylint_django/checkers/foreign_key_strings.py
@@ -1,10 +1,12 @@
 from __future__ import absolute_import
+
+import astroid
 from pylint.checkers import BaseChecker
-from pylint.interfaces import IAstroidChecker
 from pylint.checkers.utils import check_messages
+from pylint.interfaces import IAstroidChecker
+
 from pylint_django.__pkginfo__ import BASE_ID
 from pylint_django.transforms import foreignkey
-import astroid
 
 
 class ForeignKeyStringsChecker(BaseChecker):
@@ -41,15 +43,27 @@ Consider passing in an explicit Django configuration file to match your project 
         ),
     )
 
-    msgs = {"E%s10" % BASE_ID: ("Django was not configured. For more information run pylint --load-plugins=pylint_django --help-msg=django-not-configured", "django-not-configured", _LONG_MESSAGE),
-    "F%s10" % BASE_ID: (
-        'Provided Django settings %s could not be loaded',
-        'django-settings-module-not-found',
-        'The provided Django settings module %s was not found on the path'
-    )}
+    msgs = {
+        "E%s10"
+        % BASE_ID: (
+            "Django was not configured. For more information run"
+            "pylint --load-plugins=pylint_django --help-msg=django-not-configured",
+            "django-not-configured",
+            _LONG_MESSAGE,
+        ),
+        "F%s10"
+        % BASE_ID: (
+            "Provided Django settings %s could not be loaded",
+            "django-settings-module-not-found",
+            "The provided Django settings module %s was not found on the path",
+        ),
+    }
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._raise_warning = False
 
     def open(self):
-        self._raise_warning = False
         # This is a bit of a hacky workaround. pylint-django does not *require* that
         # Django is configured explicitly, and will use some basic defaults in that
         # case. However, as this is a WARNING not a FATAL, the error must be raised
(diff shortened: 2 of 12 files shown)


pylint_django/checkers/django_installed.py
```diff
@@ -20,8 +20,11 @@ class DjangoInstalledChecker(BaseChecker):
...
     @check_messages('django-not-available')
-    def close(self):
+    def open(self):
         try:
             __import__('django')
         except ImportError:
-            self.add_message('F%s01' % BASE_ID)
+            # mild hack: this error is added before any modules have been inspected
+            # so we need to set some kind of value to prevent the linter failing to it
+            self.linter.set_current_module('project global')
```
atodorov, line 29: Could be `pylint_django` but either way is fine. IMO `project global` is something that is more puzzling, since it doesn't exist, but also isn't a valid Python module name.

pylint_django/checkers/foreign_key_strings.py
```diff
@@ -0,0 +1,142 @@
...
+                    )
+                    # however we'll trundle on with basic settings
+                    from django.conf import (  # pylint: disable=import-outside-toplevel
+                        settings,
+                    )
+
+                    settings.configure()
+                    django.setup()
+
+        # now we can add the trasforms speciifc to this checker
```
atodorov, line 130: typo -> specific

pylint_django/checkers/migrations.py
```diff
@@ -156,7 +156,7 @@ def func(apps, schema_editor):
     return is_migrations_module(node.parent)
 
 
-def load_configuration(linter):
+def load_configuration(linter):  # TODO this is redundant and can be  removed
```
atodorov, line 159: +1 for adding the TODOs as separate issues so we don't forget about them. Interestingly enough the pylint test job in Travis should be catching this.

pylint_django/transforms/__init__.py
```diff
@@ -1,13 +1,20 @@
...
 import os
 import re
 
 import astroid
 
-from pylint_django.transforms import foreignkey, fields
+from pylint_django.transforms import fields
 
-
-foreignkey.add_transform(astroid.MANAGER)
```
atodorov, line 10: Could benefit from a comment either here or on the doc-string text and/or the foreignkey module to explicitly state why `foreignkey.add_transform()` isn't called like for the other modules.

pylint_django/transforms/foreignkey.py
```diff
@@ -108,10 +108,10 @@ def infer_key_classes(node, context=None):
                     # 'auth.models', 'User' which works nicely with the `endswith()`
                     # comparison below
                     module_name += '.models'
-                except ImproperlyConfigured:
+                except ImproperlyConfigured as exep:
                     raise RuntimeError("DJANGO_SETTINGS_MODULE required for resolving ForeignKey "
                                        "string references, see Usage section in README at "
-                                       "https://pypi.org/project/pylint-django/!")
+                                       "https://pypi.org/project/pylint-django/!") from exep
```
atodorov, line 114: I learned something new today, thanks! I don't even see this documented in the Python docs.

scripts/build.sh
```diff
@@ -69,7 +69,7 @@ echo "..... Trying to install the new tarball inside a virtualenv"
 # note: installs with the optional dependency to verify this is still working
 virtualenv .venv/test-tarball
 source .venv/test-tarball/bin/activate
-pip install --no-binary :all: -f dist/ pylint_django[with_django]
+pip install -f dist/ pylint_django[with_django]
```
atodorov, line 72: The reason for this line & the reason why it is failing is the same - it tries installing pylint-django and all of its dependencies from tartballs. One of them is either new or stopped shipping tarballs, and only has wheel packages. See 10 lines below for how we deal with another package which doesn't ship wheels.

My preferred way for dealing with this is to find which package is the one not providing tarballs and
1) Open an issue against it
2) Install it explicitly by itself.

atodorov (commented):
Overall looks good to me, just a couple of nit picks and comments.

pylint_django/transforms/foreignkey.py
carlio: I also only just learned about it, because pylint raised a warning in our CI linting :-)

pylint_django/checkers/django_installed.py
carlio: Agreed, `pylint_django` does make more sense. I'll change this.

pylint_django/checkers/migrations.py
carlio: Agreed - I had initially not bothered as I had planned to immediately address them in a new PR but I got distracted / didn't get time. Therefore I've made https://github.com/PyCQA/pylint-django/issues/289 and https://github.com/PyCQA/pylint-django/issues/288

pylint_django/transforms/__init__.py
carlio: There is a module-level docstring explaining this already which I think is sufficient: https://github.com/PyCQA/pylint-django/blob/feature/improve-django-settings-handling/pylint_django/transforms/__init__.py#L1

scripts/build.sh
carlio: For now I'll accept this in our own CI while I track down which dependency is the guilty party - I am fairly sure it's clikit and I'll raise a PR/issue there if that's true or as you suggest, explicitly install it ourselves.

pylint_django/transforms/__init__.py
carlio: I also added this issue as a discussion for moving all transforms/augmentations into a checker rather than being loaded immediately by the plugin: https://github.com/PyCQA/pylint-django/issues/290

scripts/build.sh
carlio: I created a new issue to separate this out: https://github.com/PyCQA/pylint-django/issues/291

Issue #288, opened 2020-08-28
Remove previous python2 related code


carlio:
There are a few Python 2vs3 related compatability statements that can be removed now that pylint_django simply doesn't support python2, so just to clean up, remove things like https://github.com/PyCQA/pylint-django/blob/master/pylint_django/utils.py#L11 and the uft-8 encoding declarations

PR #292, opened 2020-08-28
Remove old Python2 compatability code and bump release major version

carlio:
This is addressing #288 

There are several conditionals relating to Python 2 compatibility that are no longer necessary now that latest pylint, pylint-django and Django do not support python2. Additionally, Django 3 removed compatibility APIs too  (see https://docs.djangoproject.com/en/3.1/releases/3.0/#removed-private-python-2-compatibility-apis)

This PR cleans up all of those things to remove code that (as far as I know) isn't useful to keep around.

As a side-effect, I suggest bumping the pylint-django major version to match the Django one, since this likely has a few backwards-incompatible changes (probably a few extra warnings about `__str__` and `__unicode__` might appear for some users). If you disagree I'm happy to remove that part of this change!

### DIFF
--- a/pylint_django/checkers/models.py
+++ b/pylint_django/checkers/models.py
@@ -4,8 +4,8 @@ from astroid.nodes import Assign
 from astroid.nodes import ClassDef, FunctionDef, AssignName
 
 from pylint.interfaces import IAstroidChecker
-from pylint.checkers.utils import check_messages
 from pylint.checkers import BaseChecker
+from pylint.checkers.utils import check_messages
 
 from pylint_django.__pkginfo__ import BASE_ID
 from pylint_django.utils import node_is_subclass
@@ -15,13 +15,13 @@ MESSAGES = {
     "E%d01"
     % BASE_ID: (
         "__str__ on a model must be callable (%s)",
-        "model-unicode-not-callable",
+        "model-str-not-callable",
         "Django models require a callable __str__ method",
     ),
     "W%d01"
     % BASE_ID: (
         "No __str__ method on model (%s)",
-        "model-missing-unicode",
+        "model-missing-str",
         "Django models should implement a __str__ method for string representation "
         "(see https://docs.djangoproject.com/en/3.1/ref/models/instances/#django.db.models.Model.__str__)",
     ),
@@ -35,7 +35,7 @@ MESSAGES = {
     "W%d03"
     % BASE_ID: (
         "Model does not explicitly define __str__ (%s)",
-        "model-no-explicit-unicode",
+        "model-no-explicit-str",
         "Django models should implement a __str__ method for string representation. "
         "A parent class of this model does, but ideally all models should be explicit.",
     ),
@@ -57,29 +57,6 @@ def _is_meta_with_abstract(node):
             if meta_child.value.value:
                 # this class is abstract
                 return True
-    return False
-
-
-def _has_python_2_unicode_compatible_decorator(node):
-    if node.decorators is None:
-        return False
-
-    for decorator in node.decorators.nodes:
-        if getattr(decorator, "name", None) == "python_2_unicode_compatible":
-            return True
-
-    return False
-
-
-def _is_unicode_or_str_in_python_2_compatibility(method):
-    if method.name == "__unicode__":
-        return True
-
-    if method.name == "__str__" and _has_python_2_unicode_compatible_decorator(
-        method.parent
-    ):
-        return True
-
     return False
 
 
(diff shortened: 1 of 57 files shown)


carlio:
@atodorov What do you think about moving to a 3.x version?

atodorov:
> @atodorov What do you think about moving to a 3.x version?

We discussed this in https://github.com/PyCQA/pylint-django/pull/283. 

I prefer following the major version of pylint. OTOH there will be time when both pylint and Django will likely introduce some backwards incompatibilities and we'll be in the same situation again. Maybe the right course of action will be to forget about keeping versions in sync with anyone else and just increase them as necessary.  Yet I'm sure this will get people confused. I don't see a good choice unless we start keeping several active branches so may as well start doing whatever we want with the version number. A timestamp version would be an ugly but effective compromise.

tox.ini
```diff
@@ -17,7 +17,7 @@ commands =
     django_not_installed: bash -c \'pylint --load-plugins=pylint_django setup.py | grep django-not-available\'
     django_is_installed: pylint --load-plugins=pylint_django --disable=E5110 setup.py
     flake8: flake8
-    pylint: pylint --rcfile=tox.ini -d missing-docstring,too-many-branches,too-many-return-statements,too-many-ancestors,fixme --ignore=tests pylint_django setup
+    pylint: pylint --rcfile=tox.ini -d missing-docstring,too-many-branches,too-many-return-statements,too-many-ancestors,fixme,duplicate-code --ignore=tests pylint_django setup
```
atodorov: FTR I find duplicate-code checker to be too relaxed. It will match any 3 lines which are the same/similar and will produce lots of false positives. +1 for disabling.

pylint_django/checkers/models.py
```diff
@@ -35,7 +35,7 @@
     "W%d03"
     % BASE_ID: (
         "Model does not explicitly define __str__ (%s)",
-        "model-no-explicit-unicode",
+        "model-no-explicit-str",
```
atodorov, line 38: IMO message-id changes need to be documented in the changelog b/c people rely on them to enable and disable checkers. For someone this will suddenly result in lots of warnings in their pylint output.

pylint_django/tests/input/external_django_tables2_noerror_meta_class.py
```diff
@@ -2,7 +2,7 @@
 # don't produce old-style-class warnings, see
 # https://github.com/PyCQA/pylint-django/issues/56
 
-# pylint: disable=missing-docstring,too-few-public-methods
+# pylint: disable=model-no-explicit-str,model-no-explicit-str,missing-docstring,too-few-public-methods
```
atodorov, line 5: Some of the files have this repeated twice! Look below.  This could also be the cause for flake8 complaining.

atodorov (commented):
The changes are generally OK but it looks like you've been in a hury. See my other comment about the versioning scheme.

tox.ini
carlio: Yeah the issue here was with imports in two checkers being the same in the same order, there's no way round that (some base class where you import * from the parent is also bad) so it was a bit annoying as `# pylint: disable=` didn't work either. Shame but never mind.

pylint_django/checkers/models.py
carlio: I did think about that but then I forgot about it; on the one hand the message code should have meaning but as you say it'd mess up a bunch of config. I'm not sure how to handle that.

pylint_django/tests/input/external_django_tables2_noerror_meta_class.py
carlio: Ah I thought I caught these already, my `sed` to insert these went wrong a few times and I thought I backed out the changes but seems I have missed a couple.

carlio:
If the versioning follows either, I'd prefer it to follow Django because, for example, this PR would remove compatibility with Django 2. 

Having said that, this was a speculative PR - I'm aware I've been on the sidelines for several years so wanted to put some PRs and feelers out since you've been the main-maintainer for a long time! 

I always wanted to keep as much backwards-compatibility as possible but I remember discussions where you didn't like the `compat.py` but I think I went to far the other direction with this PR :-)

pylint_django/checkers/models.py
atodorov: Numeric message codes are deprecated in Pylint. They always tell you to use the human readable ones. IMO a simple conversion table `old-mesage` -> `new-message` in Changelog is enough. 

Another, more complicated option is to keep the old message ids and then add new ones for the same messages to enable a transition period. Not worth it I think.

atodorov: >  I'm not sure how to handle that

Revisiting this PR. Not sure if we can have multiple message codes for the same error.

carlio:
Closing this as it's super old and out of date now!

PR #293, opened 2020-08-28
Remove unecessary migrations blacklisting

carlio:
From what I can see, the process is:

1) pylint loads the pylint_django plugin and runs `load_configuration` (or, using compat, we do it) and we add `migrations` to the `linter.config.blacklist` : https://github.com/PyCQA/pylint-django/blob/7721691d73a26fea3d4c7e23b90d5cc8ad9c6f96/pylint_django/plugin.py#L13
2) pylint_django loads its list of checkers including the migrations checker, which removess `migrations` from `linter.config.blacklist` : https://github.com/PyCQA/pylint-django/blob/7721691d73a26fea3d4c7e23b90d5cc8ad9c6f96/pylint_django/checkers/migrations.py#L159

So basically we add it to the blacklist then remove it almost immediately. I think from the comment that the migrations checker is trying to only remove it for its own checks, but the linter config is global so I believe that this code is unused and has no effect.

Since it probably also interferes with any user-provided configuration in their `.pylintrc`, and as far as I can tell it hasn't caused any problems that everything checks `migrations` folders, I think we should remove the code to clean up as I think it's essentially a no-op

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,8 @@ Version 2.3.1 (TBD)
   fatal error behaviour. `#277 <https://github.com/PyCAQ/pylint-django/issues/277>`__
   and `#243 <https://github.com/PyCAQ/pylint-django/issues/243>`__
 - Fixed tests to work with pylint>2.6
+- Cleaned up some redundant (and possibly unused) blacklist management surrounding
+  linting of DB migrations `#289 <https://github.com/PyCQA/pylint-django/issues/289>_`
 
 Version 2.3.0 (05 Aug 2020)
 ---------------------------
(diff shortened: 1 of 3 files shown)


atodorov:
> So basically we add it to the blacklist then remove it almost immediately. I think from the comment that the migrations checker is trying to only remove it for its own checks, but the linter config is global so I believe that this code is unused and has no effect.

Migrations checker is not enabled by default. You need to `--load-plugins=pylint_django.checkers.migrations`

carlio:
@atodorov Ah I didn't know that the migrations was a separate plugin, hence my confusion. In that case I agree, this PR is unecessary as the behaviour is currently correct.

PR #300, opened 2021-01-06
Pin Faker to Prevent Asteroid Crash and Fix CI

jpulec:
There's some sort of issue with version 5.0.2 of Faker which updates a
lot of code to use the pathlib library instead of os utils. This causes
the sys.path_importer_cache to get filled with PosixPath objects instead
of strings as it's keys, which causes an asteroid crash.

For now, we can just pin Faker to a version that still works until the
upstream issue is resolved.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,7 +32,6 @@ setup(
         'Operating System :: Unix',
         'Topic :: Software Development :: Quality Assurance',
         'Programming Language :: Python :: 3',
-        'Programming Language :: Python :: 3.5',
         'Programming Language :: Python :: 3.6',
         'Programming Language :: Python :: 3.7',
         'Programming Language :: Python :: 3.8',
--- a/setup.py
+++ b/setup.py
@@ -19,6 +19,7 @@ setup(
     install_requires=[
         'pylint-plugin-utils>=0.5',
         'pylint>=2.0',
+        'Faker==5.0.1',
     ],
     extras_require={
         'with_django': ['Django'],
--- a/tox.ini
+++ b/tox.ini
@@ -9,7 +9,7 @@ envlist =
     pylint
     readme
     py{36}-django{111,20,-master}
-    py{35,36,37}-django22
+    py{36,37}-django22
     py{36,37,38}-django{30,31}
 
 [testenv]
@@ -20,7 +20,7 @@ commands =
     pylint: pylint --rcfile=tox.ini -d missing-docstring,too-many-branches,too-many-return-statements,too-many-ancestors,fixme --ignore=tests pylint_django setup
     readme: bash -c \'python setup.py -q sdist && twine check dist/*\'
     py{36}-django{111,20,-master}: coverage run pylint_django/tests/test_func.py -v
-    py{35,36,37}-django22: coverage run pylint_django/tests/test_func.py -v
+    py{36,37}-django22: coverage run pylint_django/tests/test_func.py -v
     py{36,37,38}-django{30,31}: coverage run pylint_django/tests/test_func.py -v
     clean: find . -type f -name '*.pyc' -delete
     clean: find . -type d -name __pycache__ -delete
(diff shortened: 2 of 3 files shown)


setup.py
```diff
@@ -19,6 +19,7 @@
     install_requires=[
         'pylint-plugin-utils>=0.5',
         'pylint>=2.0',
+        'Faker==5.0.1',
```
atodorov, line 22: Thanks for digging into this. Can you also remove the test jobs for Python 3.5 because Faker includes:
`python_requires=">=3.6"` and that causes these jobs to fail anyway.

atodorov: Another question: Does v 5.0.2 or any of the others work or this is the last one which works fine ?

jpulec: Just removed python 3.5 from the tests.  Faker 5.0.1 appears to be the last version that works. [This PR](https://github.com/joke2k/faker/pull/1339/files) which landed in v5.0.2 is the one that is causing issues.

atodorov (approved):

tox.ini
```diff
@@ -47,7 +47,7 @@ whitelist_externals =
     django_not_installed: bash
     readme: bash
     py{36}-django{111,20,-master}: coverage
-    py{35,36,37}-django22: coverage
+    py{36,37}-django22: coverage
```
atodorov, line 50: Also need to remove the job entries from `.travis.yml`

atodorov (changes requested):

tox.ini
atodorov: Python version that is.

atodorov:
> There's some sort of issue with version 5.0.2 of Faker which updates a
> lot of code to use the pathlib library instead of os utils. This causes
> the sys.path_importer_cache to get filled with PosixPath objects instead
> of strings as it's keys, which causes an asteroid crash.
> 

I'm not immediately finding Faker as a dependency in astroid code base. If you can share some pointers files/methods you looked at that would be helpful.


> For now, we can just pin Faker to a version that still works until the
> upstream issue is resolved.

Seems to me this isn't reported upstream so I want to gather as much info as possible before looking more into it.

atodorov (approved):

jpulec:
> I'm not immediately finding Faker as a dependency in astroid code base. If you can share some pointers files/methods you looked at that would be helpful.

Sure. Asteroid doesn't depend on Faker, but due to some global import state it causes this issue. If you check out [this build log](https://travis-ci.org/github/PyCQA/pylint-django/jobs/750055276#L447) from the first time the travis builds failed, you'll see the actual error is:

```python
AttributeError: 'PosixPath' object has no attribute 'rstrip'
```

The first `astroid` (and not `importlib`) stack frame where this happens is in `astroid/interpreter/_import/spec.py:281`.

Here's the surrounding relevant code:

```python
def _search_zip(modpath, pic):
    for filepath, importer in list(pic.items()):
        if importer is not None:
            found = importer.find_module(modpath[0])
```

If you look further up the stack, you'll find that `ZipFinder` is the class causing issues. In the [`find_module`](https://github.com/PyCQA/astroid/blob/2d25e84587c3e392751280490355aaeda7afd116/astroid/interpreter/_import/spec.py#L192) method, which passes `self._zipimporters` for the `pic` parameter of `_search_zip`.

`self._zipimporters` is populated by [`_precache_zipimporters`](https://github.com/PyCQA/astroid/blob/2d25e84587c3e392751280490355aaeda7afd116/astroid/interpreter/_import/spec.py#L259), which references `sys.path_importer_cache`.

I'm not exactly sure how/why, but somehow Faker v5.0.2+ is adding entries to `sys.path_importer_cache` which are `pathlib.PosixPath` objects instead of actual `str`s. Eventually this results in `FileFinder` instance trying to call `find_module()` with a `PosixPath` instead of a `str` which is what yields the initial error.

Hopefully that can provide a little more color. Let me know if any of that is unclear.

Issue #302, opened 2021-01-06
pinned Faker==5.0.1 causing packages installation conflicts


robinchow:
Due to this update (https://github.com/PyCQA/pylint-django/commit/9b50595800fb9f5b2063db2b9c391024ab39e4c6) and the pin being so restrictive, we are encountering the problem of "incompatible versions in resolved dependencies" when installing packages in our environment (we use `Faker==4.18.0`)

If you're trying to avoid Faker version 5.0.2, maybe instead use the requirement `Faker<5.0.2` ?

atodorov:
@robinchow makes sense. Do you mind sending a PR ?

carlio:
This was fixed as part of #304 and #305, it was just released as version 2.4.2

PR #303, opened 2021-01-07
Relaxed Faker package versioning requirement

robinchow:
As per https://github.com/PyCQA/pylint-django/issues/302. This still avoids Faker version 5.0.2, but will better avoid dependencies clashes with projects that uses pylint-django along with Faker.


atodorov:
Thanks. Pushed as 2.4.1 to PyPI.

PR #313, opened 2021-03-05
Include 'all' as a field attribute of ManyToMany relationships.

YPCrumble:
Thanks for maintaining this great repo! This is really an issue, but the fix is easy so thought it better to post as a PR. If I've missed something feel free to close. 

Per [the docs](https://docs.djangoproject.com/en/3.1/topics/db/examples/many_to_many/) It's possible to call `all()` on a many to many attribute. For example, the docs call `a1.publications.all()` in their example.

I'm seeing this error pop up which I don't believe is a configuration error.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+
+Version 2.4.3 (05 Mar 2021)
+---------------------------
+
+- Added `all` to available ManyToMany field attributes. (@ypcrumble)
+
+
 Version 2.4.2 (08 Jan 2021)
 ---------------------------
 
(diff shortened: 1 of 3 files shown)


atodorov (changes requested):
Change looks good but you need to add the tests for it as well.

YPCrumble:
@atodorov thank you! Added the test - I believe this is correct but let me know if it's not? Also added to the CHANGELOG for release today but happy to remove or edit that if you'd prefer.

YPCrumble:
@atodorov I'm actually going to close this for now - I'm not 100% sure it's necessary as this is actually an issue with my factories, not a plain Django model.

PR #317, opened 2021-03-30
Add all default methods to view subclass validation function

naquiroz:
This pull request adds all the Django's **View** class default HTTP methods to the `is_model_view_subclass_method_shouldnt_be_function`.


atodorov:
PR looks good but there are too many unrelated failures. Let me try fixing those up a bit before we can merge this. I want to see CI reporting PASS before pressing the button.

PR #319, opened 2021-04-08
Fix tests and also run them against Python 3.9 & Django 3.2 LTS

michael-k:
Closes PyCQA/pylint-django#318


atodorov:
Many thanks for taking care of this.

PR #344, opened 2022-01-02
Astroid 2.9.1 scoped_nodes compatibility fix

### LINKED ISSUE #343: Plugin appears to broken by astroid 2.9.1
With `astroid` 2.9.1, `pylint-django` appears not to work correctly.

For example, it produces errors like `Class 'MyModel' has no 'objects' member (no-member)`, even when it should be able to infer that the `objects` member exists.

Full repro steps, including `Pipenv.lock` file with precise package versions, is available in the following repo:

https://github.com/amdw/pylint-django-repro

carlio:
`astroid.scoped_nodes` is deprecated and will be moved to `astroid.nodes.scoped_nodes` in astroid 3.0 - see https://github.com/PyCQA/astroid/blob/main/astroid/scoped_nodes.py#L26

However it seems like the release of `2.9.1` moved some members already, so this PR just uses the new locations.

Fixes #343

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,10 +3,19 @@ Changelog
 
 Version 2.5 [WIP]
 ---------------------------
+
+Bugfixes
+~~~~~~~~
+
+- Fixed compatability with astroid 2.9.1 (`#343 <https://github.com/PyCQA/pylint-django/issues/343>`_)
+
+New
+~~~
+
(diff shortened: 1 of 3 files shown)


carlio:
ping @atodorov FYI

The tests all pass so I'm happy to merge it directly, but letting you know here for visibility.

PR #365, opened 2022-07-07
Suppress too-many-ancestors for ModelViewSets

jhhayashi:


atodorov (changes requested):
Overall looks good but needs tests. 

1) see `extras_require` in `setup.py` and add django-rest-framework as dependency
2) See the `pylint_django/tests/input/external_*` files and based on the same pattern add a ModelViewSet definition which we expect to not trigger too-may-ancestors error messages.

PR #372, opened 2022-09-21
Add "pk" to the good_names list

simonpanay:
As "qs", maybe "pk" could be added to the good_names list.
pk being a default attribute in every model in Django

If a test must be written for this, I am willing to to so

### DIFF
--- a/pylint_django/plugin.py
+++ b/pylint_django/plugin.py
@@ -15,6 +15,7 @@ def load_configuration(linter):
     """
     name_checker = get_checker(linter, NameChecker)
     name_checker.config.good_names += (
+        "pk",
         "qs",
         "urlpatterns",
         "register",
--- a/pylint_django/tests/input/func_noerror_model_unicode_lambda.py
+++ b/pylint_django/tests/input/func_noerror_model_unicode_lambda.py
@@ -1,7 +1,7 @@
 """
 Ensures that django models without a __unicode__ method are flagged
 """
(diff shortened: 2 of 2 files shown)


atodorov:
> If a test must be written for this, I am willing to to so

PR looks good to me but I'd hate to merge with failing tests. Do you mind patching them up ? I see a `11: unnecessary-lambda-assignment` warning which should be easy to mark as ignored. Not sure about the `django not installed` test job though.

simonpanay:
I got rid of the `unnecessary-lambda-assignment` warning, but there is still a `AttributeError: Can't pickle local object 'ArgumentParser.__init__.<locals>.identity'` thrown in `test_linter_should_be_pickleable_with_pylint_django_plugin_installed`
And I have no idea how to solve this one.
No clue how to fix the `django not installed` too.
Do you have any advice on these two ?

atodorov:
@simonpanay thanks a bunch for the fix. I will go ahead and merge this PR regardless of the failures. 

I'm not sure what's going on and I can see that some CI jobs have been failing for a while and right now (next 3-4 weeks) I am too preoccupied with personal items to be able to even debug this.

I'll happily merge any other PRs that are fixing tests if you would like to dig deeper. 

@carlio - I think we should make these CI jobs required so that we don't get around with merging on red. Otherwise there isn't much value in having tests if we can't fully trust them or if we are going to ignore the failures anyway. WDT ?

atodorov (approved):

PR #373, opened 2022-09-22
Migrate metadata from setup.py to setup.cfg

cclauss:
Done using https://pypi.org/project/setuptools-py2cfg plus manual modifications.

After this is reviewed and landed, a migration to `pyproject.toml` can be achieved using https://pypi.org/project/ini2toml or similar.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,45 @@
+[metadata]
+name = pylint-django
+version = 2.5.3
+author = landscape.io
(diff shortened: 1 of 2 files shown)


atodorov (changes requested):
Why is this change needed?

carlio:
I have been (manually) moving over to poetry so I will close this as basically done

Issue #405, opened 2023-10-01
Compatibility with pylint 3.0


matejsp:
When trying to test upgrade pylint to 3.0 that will be released in couple of days.

```
pip install pylint==3.0.0b0
Requirement already satisfied: platformdirs>=2.2.0 in ./.venv/lib/python3.11/site-packages (from pylint==3.0.0b0) (3.10.0)
Requirement already satisfied: isort<6,>=4.2.5 in ./.venv/lib/python3.11/site-packages (from pylint==3.0.0b0) (5.12.0)
Requirement already satisfied: mccabe<0.8,>=0.6 in ./.venv/lib/python3.11/site-packages (from pylint==3.0.0b0) (0.7.0)
Requirement already satisfied: tomlkit>=0.10.1 in ./.venv/lib/python3.11/site-packages (from pylint==3.0.0b0) (0.12.1)
Requirement already satisfied: dill>=0.3.6 in ./.venv/lib/python3.11/site-packages (from pylint==3.0.0b0) (0.3.6)
Installing collected packages: astroid, pylint
  Attempting uninstall: astroid
    Found existing installation: astroid 2.15.6
    Uninstalling astroid-2.15.6:
      Successfully uninstalled astroid-2.15.6
  Attempting uninstall: pylint
    Found existing installation: pylint 2.17.5
    Uninstalling pylint-2.17.5:
      Successfully uninstalled pylint-2.17.5
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
pylint-django 2.5.3 requires pylint<3,>=2.0, but you have pylint 3.0.0b0 which is incompatible.
Successfully installed astroid-3.0.0 pylint-3.0.0b0
```

```
(.venv) ➜  xxxx pylint .
Traceback (most recent call last):
  File "/Users/myuser/projects/xxxx/.venv/bin/pylint", line 8, in <module>
    sys.exit(run_pylint())
             ^^^^^^^^^^^^
  File "/Users/myuser/projects/xxxx/.venv/lib/python3.11/site-packages/pylint/__init__.py", line 34, in run_pylint
    PylintRun(argv or sys.argv[1:])
  File "/Users/myuser/projects/xxxx/.venv/lib/python3.11/site-packages/pylint/lint/run.py", line 162, in __init__
    args = _config_initialization(
           ^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #407, opened 2023-10-06
Support for pylint 3.x

matejsp:
Fixing https://github.com/pylint-dev/pylint-django/issues/405

### DIFF
--- a/pylint_django/checkers/foreign_key_strings.py
+++ b/pylint_django/checkers/foreign_key_strings.py
@@ -2,10 +2,9 @@ from __future__ import absolute_import
 
 import astroid
 from pylint.checkers import BaseChecker
-from pylint.checkers.utils import check_messages
-from pylint.interfaces import IAstroidChecker
 
 from pylint_django.__pkginfo__ import BASE_ID
+from pylint_django.compat import check_messages
 from pylint_django.transforms import foreignkey
 
 
@@ -26,8 +25,6 @@ This can be done via the DJANGO_SETTINGS_MODULE environment variable or the pyli
 
 Some basic default settings were used, however this will lead to less accurate linting.
 Consider passing in an explicit Django configuration file to match your project to improve accuracy."""
-
-    __implements__ = (IAstroidChecker,)
 
     name = "Django foreign keys referenced by strings"
 
@@ -95,7 +92,12 @@ Consider passing in an explicit Django configuration file to match your project 
             # this means that Django wasn't able to configure itself using some defaults
             # provided (likely in a DJANGO_SETTINGS_MODULE environment variable)
             # so see if the user has specified a pylint option
-            if self.config.django_settings_module is None:
+            if hasattr(self, "linter"):
+                django_settings_module = self.linter.config.django_settings_module
+            else:
+                django_settings_module = self.config.django_settings_module
+
+            if django_settings_module is None:
                 # we will warn the user that they haven't actually configured Django themselves
                 self._raise_warning = True
                 # but use django defaults then...
@@ -108,7 +110,7 @@ Consider passing in an explicit Django configuration file to match your project 
                 try:
                     from django.conf import Settings, settings  # pylint: disable=import-outside-toplevel
 
-                    settings.configure(Settings(self.config.django_settings_module))
+                    settings.configure(Settings(django_settings_module))
                     django.setup()
                 except ImportError:
                     # we could not find the provided settings module...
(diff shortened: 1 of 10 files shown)


Pierre-Sassoulas (approved):
You went out of your way to support both pylint 2 and 3 :ok_hand:

carlio:
> You went out of your way to support both pylint 2 and 3 👌

This is something I have been clinging onto for a long time : new dependencies break the library, but dropping support for old dependencies breaks so many peoples' CI pipelines because they've pinned older versions. It has to be a careful mix of both  or people will just drop pylint-django (and prospector has this philosophy too).

I'm reluctant to say "this only supports the latest and greatest mix of Python+Django+Pylint" because that rules out the majority of people and companies who don't have the time to upgrade their dependencies, and honestly they're the ones who need CI static analysis the most in my opinion, to prevent building up too much technical debt since they don't have the capacity to go back and deal with it later.

I'm probably biased because a lot of my work is in helping people deal with technical debt caused by prioritizing new features and growth over maintenance.

Anyway that's my answer to your question on #405 @matejsp : "I didn't prepare PR because I don't know if you want to preserve backwards compatibility or just support pylint 3.x"

Sorry for being slow about this but I don't have much time right now, I'll get to it!

Pierre-Sassoulas:
> I'm reluctant to say "this only supports the latest and greatest mix of Python+Django+Pylint" even though that's easier for me, because that rules out the majority of people and companies who don't have the time to upgrade their dependencies, and honestly they're the ones who need CI static analysis the most in my opinion, to prevent building up too much technical debt since they don't have the capacity to go back and deal with it later.

Personally, I'm following the system, the language, or the framework's EOL guidelines (debian/python/django) witohut feeling guilty about it. If upgrading would be very costly for a company, then they can pay you a fraction of the upgrade cost so you'll maintain your code for longer than the EOL (never seen that done for a lib, but I guess Ubuntu can do it with financial success with their Expanded Security Maintenance). Everyone can use the existing old version from 4 year ago that is compatible with their old stack (and ignore the CVEs the same way they ignore the OS/language/framework CVEs).  For this reason it's even arguably not that problematic to support for less than the EOL as long as one version of your lib is compatible with interpreters that are still not EOL (i.e. pylint-django could drop support for python 3.8 / pylint 2.17.7 right now, there's still usable versions for python 3.8 / pylint 2.17.7). To be frank, if deploying prod code on an OS that is affected by heart-bleed in 2020+ is not enough of a reason for a company to stop prioritizing business above all else and finally upgrade, then I don't know why this company should survive on your back. There's no reason you in particular should suffer to allow them to survive without ripping any of the benefits.

That said it's nice to stay compatible when the cost is low or if (hypothetically) you're compensated for your work.

pylint_django/checkers/foreign_key_strings.py
```diff
@@ -95,7 +92,12 @@ def open(self):
             # this means that Django wasn't able to configure itself using some defaults
             # provided (likely in a DJANGO_SETTINGS_MODULE environment variable)
             # so see if the user has specified a pylint option
-            if self.config.django_settings_module is None:
+            if hasattr(self, "linter"):
+                django_settings_module = self.linter.config.django_settings_module
```
twanwv, line 96: This line is hit at my end (`self.config` is `None`) meaning that at line 113 the `self.config.django_settings_module` is resulting in an error, I think that 113 should be: `settings.configure(Settings(django_settings_module))` which worked for me.

Pierre-Sassoulas (commented):
I've tested locally and I confirm it works for me.

``pylint --version``
```
pylint 3.0.1
astroid 3.0.0
Python 3.10.12 (main, Jun 11 2023, 05:26:28) [GCC 11.4.0]
```

```
pip3 freeze|grep pylint-django
pylint-django @ git+https://github.com/matejsp/pylint-django.git@bab54286b0d6d5a06ffec28198a61d7ebf6a38b4
```

After fixing an [unrelated hack I did before based on a stackoverflow answer](https://stackoverflow.com/a/74242848/2519059) that "just worked":
```diff
-init-hook="import sys; import os; from pylint.config import find_pylintrc; sys.path.append(os.path.dirname(find_pylintrc()))"
+init-hook="import os;os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'toolkit.settings')"
```

carlio:
I have been trying to get the CI working (see https://github.com/pylint-dev/pylint-django/pull/400#issuecomment-1773980016) but it's a bit of a pain so I'm going to merge this and make a bugfix release first to free up peoples' own CI builds.

carlio:
Finally managed to make a bugfix release to include this. `2.5.4` is on PyPI. Thanks again @matejsp :-)

I'll get on to cleaning up the pylint-django CI process next.

PR #414, opened 2023-10-30
Migrate from isort and flake8 to ruff

Pierre-Sassoulas:
Blocked by #412 #413. Also need some workflow changes.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,21 +10,17 @@ repos:
       - id: mixed-line-ending
         args: [--fix=lf]
       - id: debug-statements
-  - repo: https://github.com/PyCQA/flake8
-    rev: 6.1.0
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: "v0.1.4"
     hooks:
-      - id: flake8
-        args: [--max-line-length=120]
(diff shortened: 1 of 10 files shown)


.pre-commit-config.yaml
```diff
@@ -8,19 +8,14 @@ repos:
...
-        args: [ --max-line-length=120 ]
+      - id: ruff
+        args: ["--fix"]
+        exclude: "tests/input/"
   - repo: https://github.com/psf/black
     rev: 23.10.1
     hooks:
       - id: black
         args: [--safe, --line-length=120]
-  - repo: https://github.com/PyCQA/isort
```
carlio, line 22: Do ruff + black play well together for sorting imports? I remember that isort came after black due to them conflicting, however that was some time ago and likely fixed.

Pierre-Sassoulas: Yeah ruff went out after black so it supports it without config. (there's also ruffmt now, but I like black for having only one option).

carlio (approved):
Seems okay to me, depends on the other stuff being merged I guess?

Pierre-Sassoulas:
Yeah, I had to rebase on everything to make it work 👌

PR #416, opened 2023-10-30
[pre-commit] Add and apply prettier

Pierre-Sassoulas:
The prettier installation in pre-commit is painfully slow, but it's a one time thing, I think the benefits are great after that.


carlio:
I think that the slow install time might be a problem in the future, as I'd quite like the CI to run pre-commit to verify that if has been run at all (not everyone will have installed it locally first) and this could delay an already slow CI time.

I will worry about that at a future time though for now :+1: for me

carlio (approved):

Pierre-Sassoulas:
It's cached on pre-commit.ci side so it's going to be a problem only locally :)

PR #421, opened 2023-11-29
use django aliases for ranges to support psycopg 2 and 3

jsatt:
Django 4.2 introduces support for psycopg 3, but when using psycopg 3 Ranges without psycopg2 install, pyling-django raises an AstroidImportError due to `apply_type_shim` checking for `psycopg2._range`. Django 4.2 also now provides aliases to the Ranges in the appropriate library. This change uses these aliases for checking and falls back using `psycopg2` if they are not available (running Django <4.2).

### DIFF
--- a/pylint_django/transforms/fields.py
+++ b/pylint_django/transforms/fields.py
@@ -79,7 +79,10 @@ def apply_type_shim(cls, _context=None):
     elif cls.name in ("HStoreField", "JSONField"):
         base_nodes = scoped_nodes.builtin_lookup("dict")
     elif cls.name in _RANGE_FIELDS:
-        base_nodes = MANAGER.ast_from_module_name("psycopg2._range").lookup("Range")
+        try:
+            base_nodes = MANAGER.ast_from_module_name("django.db.backends.postgresql.psycopg_any").lookup("Range")
+        except AstroidImportError:
+            base_nodes = MANAGER.ast_from_module_name("psycopg2._range").lookup("Range")
     else:
         return iter([cls])
 


michael-k:
I tried your branch, but now I run into a different problem.  The field is defined as `period = DateRangeField(null=False, blank=False)` and we use `period.lower`/`period.upper`.  With psycopg3, pylint complains that `Instance of 'DateRangeField' has no 'lower' member (no-member)` (same for `uppper`).

- [`Range.lower`/`.upper` in psycopg2](https://github.com/psycopg/psycopg2/blob/a971c11d5038333f08821ef452ce1237a22cdb0e/lib/_range.py#L78-L86)
- [`Range.lower`/`upper` in psycopg3](https://github.com/psycopg/psycopg/blob/558cbe7ef7ef6a6d1f6c65e368aaaf9f9a10108e/psycopg/psycopg/types/range.py#L131-L139)  (The only difference I can see is the `Generic[T]`.)

Can you please include psycop3 variants of [`external_psycopg2_noerror_postgres_fields.py`](https://github.com/pylint-dev/pylint-django/blob/a2b4459a4fa97ac6619776cb513d8120214a167c/pylint_django/tests/input/external_psycopg2_noerror_postgres_fields.py) and [`external_psycopg2_noerror_postgres_fields.rc`](https://github.com/pylint-dev/pylint-django/blob/a2b4459a4fa97ac6619776cb513d8120214a167c/pylint_django/tests/input/external_psycopg2_noerror_postgres_fields.rc) in this PR?

----

I've also tried `period[0]`/`period[1]` with psycopg3 and pylint complains that `Value 'self.period' is unsubscriptable (unsubscriptable-object)`.

- Django: 4.2.10 and 5.0.2
- psycopg: 3.1.18

carlio:
I have added the tests suggested and merged this in, to find out if they work ; if not I will fix things up. Thanks!

PR #423, opened 2023-12-12
Supress some protected-access warnings when django-simple-history is installed.

m000:
django-simple-hisory is a fairly popular [1] package for keeping track of changes in django objects. Setting the `_change_reason` property of an object is the officially documented way to provide a value for the `history_change_reason` field of historical objects [2].

When django-simple-hisory is installed, protected-access warnings for setting `_change_reason` is most likely a false positive, and should be supressed.

Because of inherent limitations of pylint, this may lead to some false negatives if `_change_reason` is used elsewhere.

[1] https://pypistats.org/packages/django-simple-history
[2] https://django-simple-history.readthedocs.io/en/latest/historical_model.html#change-reason


m000:
I know this is missing a test. I would first like to bring it up for discussion/feedback before adding one.

I am working on a codebase that has 24 cases (and counting) where the specific warning is suppressed for `_change_reason`. For this, I think this would make a meaningful addition for pylint-django.

PR #427, opened 2024-02-11
Support for python 3.12 datetime

matejsp:
Fixes https://github.com/pylint-dev/pylint-django/issues/425

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,7 @@ This version drops support for Python 3.6
 Bugfixes
 ~~~~~~~~
 
+- Fixed compatibility issue with datetime classes and python `3.12` (`#425 <https://github.com/pylint-dev/pylint-django/issues/425>`_)
 
 
 Version 2.5.3 (25 Mär 2022)
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -20,3 +20,4 @@
 - [naquiroz](https://github.com/naquiroz)
 - [john-sandall](https://github.com/john-sandall)
 - [dineshtrivedi](https://github.com/dineshtrivedi)
+- [matejsp](https://github.com/matejsp)
--- a/pylint_django/compat.py
+++ b/pylint_django/compat.py
@@ -1,6 +1,7 @@
 # flake8: noqa
 # pylint: skip-file
 # no sane linter can figure out the hackiness in this compatibility layer...
+import sys
 
 try:
     from astroid.nodes import AssignName, Attribute, ClassDef, FunctionDef, ImportFrom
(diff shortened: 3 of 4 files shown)


Pierre-Sassoulas (changes requested):
If I understand correctly the best case scenario is for python 3.12 (faster because there is no error raised then caught) ? With only 7.5% users on it, it's not the most likely interpreter for pylint-django users so it should be the other way around see https://pypistats.org/packages/pylint-django. 

daily_download_proportion_pylint_django

Also, maybe it would be better to have a "is_python_3_12" constant than catch the problem every-time ?

matejsp:
@Pierre-Sassoulas 

When creating the code I was trying to replicate the existing code with _pydecimal that uses the same try catch.

I was testing the following fix in our monolit with 500 kLOC and more than 700 db tables and around the same number of datetime fields and there wasnt any negative performance impact.

But I see your point and wish your concerns would be shared with pylint developers :D We only run on CI on AWS and in our case it takes 7 minutes with 12 core on very large ec2 instance to lint 500 kLOC code and on my computer it takes about an hour.  Pylint lack of performance and even on home page bragging how painfully slow pylint is, are making our developers cry. Otoh mypy takes about 1 minute.

I will refactor based on your proposal with a constant.

matejsp:
Pushes the changes. CI passes.

Pierre-Sassoulas (approved):
LGTM ! 

> But I see your point and wish your concerns would be shared with pylint developers :D We only run on CI on AWS and in our case it takes 7 minutes with 12 core on very large ec2 instance to lint 500 kLOC code and on my computer it takes about an hour. Pylint lack of performance and even on home page bragging how painfully slow pylint is, are making our developers cry. Otoh mypy takes about 1 minute.

As far as I remember I'm the one who wrote the readme. It should not be a bragging tone (suggestions welcome to change that btw :) ). We're playing with the cards and assumptions that were made. pylint is not using the typing and infer the typing of something by checking the values in all the calls : it's not fast, it's also hard to parallelize and we're based on the python builtin ast which has its advantages but also can't be ported to rust or C (and most contributors know python anyway). There's also organically slow check like duplicate-check.. The base problem is not an easy problem to fix, and pylint being fast is not the reason people use pylint, so we focus mainly on keeping up with python development which is already a lot for a small team of volunteers. All that to say that we do share a concern about performances and try to do our best.

Pierre-Sassoulas:
Just a heads up, I don't know when this will be released, you can use the github repo directly by using pip's git+ssh feature meanwhile.

matejsp:
Thank you for the merge. I am looking forward to the official release but will test it locally for now. On CI we would need oficial release.

PR #428, opened 2024-02-28
Actually run the tests and also run them against Django 4.1-5.0

michael-k:
When `scripts/test.sh` was introduced back in commit https://github.com/pylint-dev/pylint-django/commit/55cb670049b1d2d2219ea0968ef55f1e914e055d, [`test_func.py`](https://github.com/pylint-dev/pylint-django/blob/55cb670049b1d2d2219ea0968ef55f1e914e055d/test/test_func.py) had

```python
if __name__=='__main__':
    testlib.unittest_main(defaultTest='suite')
```

Commit https://github.com/pylint-dev/pylint-django/commit/38f2cf2e7f0909f17b6a0d67a37b89c2ffc678b0 (PR pylint-dev/pylint-django#417) removed 

```python
if __name__ == "__main__":
    sys.exit(pytest.main(sys.argv))
```

The tests on the main branch pass, but as the [logs](https://github.com/pylint-dev/pylint-django/actions/runs/8055242581) show, pytest never runs.  And the coverage shows 0%.

----

In my fork all tests for `test old stuff / Django@… / Python@3.7` don't do anything after `Resolving dependencies...` and I've cancelled the workflow after 18 minutes.  Maybe support for Python 3.7 can be dropped?  It [reached its EOL](https://peps.python.org/pep-0537/#lifespan) back in 2023-06-27.  Django 3.2 LTS is the last version that supports Python 3.7 and that will [reach its EOL in April 2024](https://www.djangoproject.com/download/).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -40,7 +40,22 @@ jobs:
       fail-fast: false
       matrix:
         python-version: [3.8, 3.9, "3.10", "3.11", "3.12"]
-        django-version: [-main, "4.0"]
+        django-version: [-main, "5.0", "4.2", "4.1", "4.0"]
+        exclude:
+          - python-version: "3.8"
+            django-version: -main
+          - python-version: "3.8"
+            django-version: "5.0"
+          - python-version: "3.9"
+            django-version: -main
+          - python-version: "3.9"
+            django-version: "5.0"
+          - python-version: "3.12"
+            django-version: "4.2"
+          - python-version: "3.12"
+            django-version: "4.1"
+          - python-version: "3.12"
+            django-version: "4.0"
 
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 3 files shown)


michael-k:
Oh, so the tests on Python 3.7 do pass, they just need a very long time.  46 minutes vs. 1-2 minutes.  I still propose to drop support for Python 3.7.  pylint 3 requires Python 3.8+.  Those who want to lint their code on Python 3.7 can still use the existing releases.

carlio:
I finally got back to dealing with this repository and cleaning it up.

I started fixing tox locally and ended up doing most of the same changes you did, I should have checked open PRs first, sorry! However the master branch now has CI which works.

I have also dropped 3,7 now, for reasons you suggested except the CI job took 6 hours (!!!) so it became even more urgent.

My general philosophy is that the people who need help from linting tools the most are people who are upgrading legacy code with from versions to newer ones, and so the tools should support that use case as much as possible. However there's a limit, and since you say, pylint no longer supports 3.7 it seems silly for pylint-django to do so.

Thanks!

PR #440, opened 2024-10-09
Add support for python 3.13

### LINKED ISSUE #432: Release version with support for Python 3.12
Heya, would it be possible to create a PyPi release with #427 included?

Pierre-Sassoulas:
Closes #432


carlio:
Just FYI I'm in UTC+7 timezone right now so I might be a bit behind when replying!

PR #461, opened 2025-05-29
Add Django 5.2 support

scpaes:
#460

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -87,6 +87,7 @@ lint.select = [
   "W",       # pycodestyle
 ]
 lint.ignore = [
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (approved):

Pierre-Sassoulas:
The fix is to rely on bleeding edge pylint only and remove the dependency to bleeding edge astroid. Bleeding edge pylint should control its dependencies. And having bleeding edge pylint be always compatible with bleeding edge astroid is something pylint is not able to guarantee in a reasonable amount of time for a reasonable amount of maintainer time. (And impossible to always be true when there's breaking changes in astroid)

Issue #467, opened 2025-10-13
Support for Pylint 4.0.0


simensol:
Hi maintainers,

Pylint 4.0.0 was released on October 12, 2025. Currently, `pylint-django` (2.6.1) requires `pylint >=3.0,<4`, which prevents using the latest Pylint version together with this plugin.

Could you please bump the dependency to allow `pylint` v4.0.0? This would help projects that want to stay up to date with the latest linting features and fixes.

Thank you for your work on this project!

Pierre-Sassoulas:
Path to pylint 4 support:
- [ ] https://github.com/pylint-dev/pylint-django/pull/440
- [ ] #472
- [ ] #468

carlio:
I personally annotate my fields a `price: Decimal = DecimalField...` for type checking in my IDE, as it tends to work better. I agree with your suspicion.

Perhaps for some combinations of python+django version, just replacing the DecimalField AST node with a Decimal node would solve the issue instead of duplicating all fields? Or is that what you already meant, if so I agree.

decibyte:
Hoping a purely supportive comment is allowed here:

I appreciate the work you're putting into adding support for pylint 4 :)

carlio:
Just released 2.7.0 with pylint 4 support

mspiller:
Latest release on pypi is missing source distribution.

PR #468, opened 2025-10-13
Support for pylint 4.0+

Pierre-Sassoulas:
Refs #467, ~blocked by #472~


Pierre-Sassoulas:
Sure (I put back 5.1), but I was just wondering as it's a minor version upgrading should be uneventful ? (as even upgrading a django major is uneventful in my experience)

PR #469, opened 2025-10-15
Update versions of Python and Django

decibyte:
I did my best at updating the dependencies of Django and Python to only be those that are not EOL (including Django 6.0, which is currently in alpha and expected to land in a couple of months).


Pierre-Sassoulas (approved):
LGTM but waiting for @carlio 's opinion on this one.

PR #474, opened 2025-11-09
Drop python 2 / pylint 2.0 remnants

Pierre-Sassoulas:
pylint-django is not compatible with pylint 2 nor python 2 anymore since some time ago, but some compat layers are still there and have a run time impact as well as a maintenance impact.

Better reviewed commit by commit.


Pierre-Sassoulas:
Thank you for merging this @carlio !

PR #483, opened 2026-07-10
Adding format-html-interpolation and mark-safe-interpolation checkers

joewesch:
This PR adds 2 checkers that revolve around the same issue: passing interpolated strings to `format_html` or `mark_safe`. When using `format_html`, it should be treated the same as an f-string and pass the values are args or kwargs so that they are automatically escaped. By using interpolated strings directly into `format_html`, you completely bypass this safety measure.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,18 @@ Changelog
 =========
 
 
+Version 2.8.0
+-------------
+
+New checks
+~~~~~~~~~~
+
+- Added ``format-html-interpolation`` (W5150) and ``mark-safe-interpolation`` (W5151) checks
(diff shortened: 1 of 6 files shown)


carlio:
This sounds like a nasty gotcha, is there a Django issue about it? It'd be good for documentation's sake to link that.

Thanks for the PR I will merge it (maybe you want to add that Django issue number as a comment) in a bit when I get home

joewesch:
@carlio, thanks for the feedback. I have added a few references.

Django deprecated calling `format_html` without args in v5 and now requires them in v6.

carlio:
sorry got distracted and forgot, doing it now :)

carlio:
release done, thanks for the PR!

